PR #449, opened 2026-02-21
448 feature port glp2 documentuploaderjsp from oscar 19

### LINKED ISSUE #448: [Feature]: Port GLP2+ documentUploader.jsp from OSCAR 19
### Problem

The existing file uses outdated libraries and liscence and clashes with the rest of the UI

### Solution

port o19

### Alternatives

_No response_

### Module/Area

Reports

### Priority

Must have

### Additional Context

Image

### Checklist

- [x] I have searched for similar feature requests
- [x] I am willing to help implement this feature

phc007:
## Description
Reworked multiple document uploader with drag and drop, multiple file, error message support
Updates license and UI to match bootstrap elsewhere in the project

## Related Issues
Fixes #448 

## How Was This Tested?
on Open O first

## Screenshots
new_Upload_Multiple_Documents

## Checklist

___

### **Description**
- This PR enhances the document uploader by updating the jQuery File Upload Plugin and related components.
- Introduces a new processing plugin for better file handling.
- Improves security measures and error handling in the upload process.
- Updates styles for better user experience and responsiveness.
- Adds new localization keys for improved internationalization support.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>jquery.fileupload.js</strong><dd><code>Update jQuery File Upload Plugin with Enhancements</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/main/webapp/share/documentUploader/jquery.fileupload.js
<li>Updated jQuery File Upload Plugin to the latest version.<br> <li> Improved file upload handling with new features and bug fixes.<br> <li> Enhanced error handling and security measures.<br>

</details>

  </td>
  <td><a href="https://github.com/carlos-emr/carlos/pull/449/files#diff-4d1a1be36f029efa365855dbd1877d0cf6113b0fbdad490d6f95990695f5f60f">+1546/-662</a></td>
</tr>

<tr>
  <td>
    <details>
      <summary><strong>jquery.tmpl.min.js</strong><dd><code>Update jQuery Template Plugin</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/main/webapp/share/documentUploader/jquery.tmpl.min.js
<li>Updated jQuery template plugin to the latest version.<br> <li> Improved template handling and performance.<br>

</details>

  </td>
  <td><a href="https://github.com/carlos-emr/carlos/pull/449/files#diff-24632a21977154bba8e208463f787b8b776904cfdfeaa001fbf9c2103c09da5f">+2/-1</a>&nbsp; &nbsp; &nbsp; </td>
</tr>

<tr>
  <td>
    <details>
      <summary><strong>jquery.iframe-transport.js</strong><dd><code>Update jQuery Iframe Transport Plugin</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/main/webapp/share/documentUploader/jquery.iframe-transport.js
<li>Updated jQuery Iframe Transport Plugin to the latest version.<br> <li> Improved handling of iframe uploads and error responses.<br>

</details>

  </td>
  <td><a href="https://github.com/carlos-emr/carlos/pull/449/files#diff-4ee09019e96a486e476574d1f212457b66efc15ba81f5df532ca1a4e7cb90a4c">+211/-140</a></td>
</tr>

<tr>
  <td>
    <details>
      <summary><strong>jquery.fileupload-process.js</strong><dd><code>Add File Upload Processing Plugin</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/main/webapp/share/documentUploader/jquery.fileupload-process.js
<li>Added jQuery File Upload Processing Plugin.<br> <li> Introduced file processing functionality to the upload workflow.<br>

</details>

  </td>
  <td><a href="https://github.com/carlos-emr/carlos/pull/449/files#diff-4234fe6f11e713e42083778d83429652f2c1e73cab8b3b496542f34058499043">+170/-0</a>&nbsp; </td>
</tr>

<tr>
  <td>
    <details>
      <summary><strong>style.css</strong><dd><code>Update Styles for File Upload</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/main/webapp/share/documentUploader/style.css
<li>Updated styles for the file upload progress bar.<br> <li> Added responsive styles for better mobile support.<br>

</details>

  </td>
  <td><a href="https://github.com/carlos-emr/carlos/pull/449/files#diff-3842c50dfae4a592c42f44444c363c56161ddf8e4f163c3c4624d571d9270b8e">+56/-20</a>&nbsp; </td>
</tr>

<tr>
  <td>
    <details>
      <summary><strong>oscarResources_en.properties</strong><dd><code>Update Localization for Document Uploader</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/main/resources/oscarResources_en.properties
<li>Added new internationalization keys for document uploader.<br> <li> Updated existing keys for better localization.<br>

</details>

  </td>
  <td><a href="https://github.com/carlos-emr/carlos/pull/449/files#diff-ec1ea2573170b49f2ce9345e4d8374ed08e9cd91bd3b610674b29a415e17d12b">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>
</tr>
</table></td></tr></tr></tbody></table>

___

> 💡 **Penify usage**:
>Comment `/help` on the PR to get a list of all available Penify tools and their descriptions

## Summary by CodeRabbit

* **New Features**
  * Modular, template-driven file upload with drag-and-drop, paste support, chunked uploads, and improved download handling
  * Per-file processing and client-side validation with real-time progress and queue management
  * Responsive Bootstrap-styled upload UI with destination/provider selection and templated file lists

* **Security**
  * CSRF token protection and HTML encoding for displayed data
  * Stricter authentication handling for access to the uploader

* **Improvements**
  * Enhanced iframe transport and broader browser/module support; new i18n strings and UI polish


phc007:
@claude 

Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Around line 446-470: Replace unsafe DOM insertion and the broad row-removal
selector: change assignments using li.innerHTML with a safe text assignment (use
textContent or equivalent) when setting values from data.result[0].error and
data.result[0].name (in the fileuploadalways handler), and narrow the
$("tr:first-child").remove() call to target the specific upload table (e.g.,
select the row within the uploader container or a specific table id/class) so
you only remove the first row of the upload table rather than every table on the
page.
- Around line 246-248: Replace hardcoded English text with i18n message keys:
change the label text "Send to Provider:" (label for="providerDrop") to use a
<fmt:message> key (e.g., document.sendToProvider), replace the option display
"None" (option value="0" in select id="providerDrop" name="providerDrop") with a
<fmt:message> key (e.g., common.none) while preserving the existing selected
expression, and likewise replace the "Queue:" label (the queue-related label in
this form) with a <fmt:message> key (e.g., document.queue) so all form labels
match the existing localization pattern.
- Line 229: Escape the context path when rendering the form action in
documentUploader.jsp: replace the unencoded "<%=context%>" used in the action
attribute of the form (documentUpload.do?method=executeUpload) with
Encode.forHtmlAttribute(context) and ensure the JSP imports the OWASP Encoder
class (org.owasp.encoder.Encode) if not already imported so
Encode.forHtmlAttribute(...) is available.
- Around line 139-150: Replace the synchronous AJAX calls in setDestination and
setDestFolder: remove async:false and convert the requests to asynchronous calls
(use jQuery.ajax without async:false or $.get/$.post) and handle success/error
callbacks to update any UI as needed (e.g., still call setDropList() after
success for setDestination if required); also URI-encode the values (use
encodeURIComponent on destination and destFolder) when building the query string
so the documentUpload.do?method=setUploadDestination and
documentUpload.do?method=setUploadIncomingDocumentFolder requests receive safe
parameters.
- Around line 227-236: The upload form with id "fileupload" that posts to
documentUpload.do is missing a CSRF hidden input; add a hidden input field
inside the form (name matching your server-side validator, e.g., "csrfToken" or
the framework-specific name) and populate its value from the server-side CSRF
provider (e.g., the request/session attribute or framework _csrf token used
elsewhere), ensuring the name matches the server's expected parameter so the
upload request will pass CSRF validation.
- Around line 301-304: There are unclosed <p> tags around the file upload markup
(the block containing the span with class "fileupload-process" and the similar
block at the other location) which produces invalid HTML; edit the
documentUploader.jsp to either remove the stray opening <p> tags or add matching
closing </p> tags so each <p> is properly closed (e.g., close the <p> that wraps
the span.fileupload-process and the other unmatched <p> at the other occurrence)
ensuring the DOM structure is valid.
- Around line 233-236: The hidden input values for destination, destFolder,
provider and queue currently inject raw JSP expressions; update the JSP to
OWASP-encode each attribute value using Encode.forHtmlAttribute() (e.g., replace
occurrences of <%=destination%>, <%=destFolder%>, <%=provider%>, and
<%=queueId%> with Encode.forHtmlAttribute(...) around the corresponding
variables) so all HTML attribute contexts use Encode.forHtmlAttribute() for safe
output in documentUploader.jsp.

In `@src/main/webapp/share/documentUploader/jquery.fileupload-ui.js`:
- Around line 552-557: The fail callback in the _editHandler is doing unsafe
property access and naive string concatenation: instead of using
data.files[index].error + data.result[0].error, guard against missing properties
by checking that data, data.files, data.files[index] and data.result and
data.result[0] exist; extract each error only if defined, then build the final
message by joining the non-empty error strings (e.g., with ' - ' or newline) or
using a fallback message; update the block around
template.find('.edit').prop('disabled', false) to compute a safe error string
and only call template.find('.error').text(...) when that safe string is
non-empty to avoid TypeError and "undefined" fragments.
- Around line 186-196: The code unsafely reads data.result[0].error and leaves a
console.log debug statement; in the done handler replace the direct access with
a guarded lookup (verify data && Array.isArray(data.result) &&
data.result.length > 0) before reading error and fall back to the existing file
default (the file variable) or a safe null/undefined error value, and remove the
console.log call so only the validated error value is used in subsequent logic
(refer to the local variables file, files, index and the data.result[0].error
access to locate the change).

---

Nitpick comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Line 466: Remove the debug console.log call; locate the AJAX success/error
handler in documentUploader.jsp where console.log(data.textStatus) is present
(the handler that receives the "data" parameter) and delete that console.log
statement (or replace it with a proper production logger call if needed),
ensuring no other logic depends on its output.

In `@src/main/webapp/share/documentUploader/jquery.fileupload.js`:
- Around line 1-604: The library relies on deprecated jQuery utilities ($.type,
$.isArray, $.inArray) that will break on jQuery 4; to address the review either
pin the project to a compatible jQuery (<3.0) in dependency manifests and
document this constraint, or add a small compatibility shim loaded before
jquery.fileupload.js that re-implements $.type, $.isArray and $.inArray (and any
other missing utilities) when they are absent so jquery.fileupload.js can
continue to call those symbols safely; locate usages of $.type, $.isArray and
$.inArray inside jquery.fileupload.js to verify coverage and include the shim
conditionally to avoid masking real upgrade issues.

In `@src/main/webapp/share/documentUploader/style.css`:
- Around line 25-30: Remove the deprecated word-wrap declaration from the .files
.name selector and rely on overflow-wrap; if you need the older break-word
fallback for legacy browsers, replace overflow-wrap: anywhere with
overflow-wrap: break-word instead of keeping word-wrap. Ensure you edit the
.files .name rule (remove word-wrap or change overflow-wrap to break-word) so
only modern, non-deprecated properties remain.

phc007:
@claude 
fix issues raised by [gemini-code-assist] that are medium or higher priority 
use the tag dms.incomingDocs.queue instead of dms.documentUploader.queue

phc007:
@yingbull 
retested (albeit on a hacked OpenO) and it works as modified
ready to review

phc007:
@claude
Verify each finding against the current code and only fix it if needed.

In `@src/main/webapp/documentManager/documentUploader.jsp` at line 72, The direct
call to Integer.parseInt(queueIdStr) can throw NumberFormatException if the
session attribute "preferredQueue" isn't numeric; update the
documentUploader.jsp to guard parsing of queueIdStr (from the preferredQueue
session value) by either wrapping Integer.parseInt(queueIdStr) in a
try/catch(NumberFormatException) and providing a safe fallback/default for
queueId, or replace it with a resilient parser like
NumberUtils.toInt(queueIdStr, defaultValue); ensure you reference
queueIdStr/queueId and handle the non-numeric case without letting the exception
propagate to a 500.

Inline comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Line 253: The value attribute currently outputs h.getProviderNo() raw; change
the JSP to encode the provider number in the HTML attribute context by wrapping
h.getProviderNo() with Encode.forHtmlAttribute() in the option value and keep
the selected comparison logic intact (i.e., use
Encode.forHtmlAttribute(h.getProviderNo()) for the value attribute while leaving
the equals(provider) check as-is or compare using the raw values before
encoding). This ensures the option value uses Encode.forHtmlAttribute() per
project guidelines.
- Around line 445-466: The code assumes data.result[0] exists; guard against
empty/non-array results by first verifying Array.isArray(data.result) &&
data.result.length > 0 (or data.result && data.result[0]) before accessing
.error or .name; if the array is empty or the first element is falsy, treat it
as an upload failure and append a generic error message to `#msg` (or handle as
your UI expects) instead of accessing properties on undefined—update the
branches that reference data.result[0].error and data.result[0].name
accordingly.

In `@src/main/webapp/share/documentUploader/jquery.fileupload-ui.js`:
- Around line 676-679: The toggle change handler on the widget isn't being
unbound because the event name passed to this._off is incorrect; update the call
that currently targets this.element.find('.fileupload-buttonbar .toggle') and
passes the event string 'change.' to instead pass 'change' (remove the trailing
dot) so the name matches the handler registered by this._on and the listener is
properly removed during widget destruction.
- Around line 85-90: getFilesFromResponse currently expects data.result.files
but the server returns a flat array (e.g., [{name,size,error}]), so update the
client to accept the flat shape: in the fileupload initializer (where you call
$('#fileupload').fileupload(...)) override the getFilesFromResponse option to
check if data.result is an Array and return it, otherwise fall back to
data.result.files or []; alternatively, standardize the server executeUpload
response to return { files: [...] } so getFilesFromResponse can remain
unchanged; locate the existing getFilesFromResponse usage and the
$('#fileupload').fileupload initializer to apply the fix.
- Line 163: The code assumes $.support.transition exists causing incorrect
behavior and potential errors; update jquery.fileupload-ui.js to safely detect
transition support (e.g., compute a local flag like supportTransition =
!!($.support && $.support.transition)) and use that flag instead of directly
referencing $.support.transition in the .addClass call and anywhere
$.support.transition.end is accessed (replace direct uses in the progress UI
creation and at the two sites that reference .end with guarded
access/conditional logic so you only add the 'progress-animated' class or read
.end when supportTransition is truthy). Ensure all three locations (the
.addClass line that currently uses !$.support.transition, and the two places
that reference $.support.transition.end) use the new guarded flag/conditional
access.
- Around line 607-633: The plugin assumes $.support.transition exists, but
jQuery 3.x removed $.support causing _forceReflow and _transition to misbehave
and skip fade handling; add a small compatibility shim executed before the
fileupload plugins load that defines $.support.transition (detecting supported
transitionend event names) so _transition and _forceReflow can rely on it—ensure
the shim runs early (before jquery.fileupload-ui.js) and references the same
$.support.transition used by the _transition method.

---

Outside diff comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Line 72: The direct call to Integer.parseInt(queueIdStr) can throw
NumberFormatException if the session attribute "preferredQueue" isn't numeric;
update the documentUploader.jsp to guard parsing of queueIdStr (from the
preferredQueue session value) by either wrapping Integer.parseInt(queueIdStr) in
a try/catch(NumberFormatException) and providing a safe fallback/default for
queueId, or replace it with a resilient parser like
NumberUtils.toInt(queueIdStr, defaultValue); ensure you reference
queueIdStr/queueId and handle the non-numeric case without letting the exception
propagate to a 500.

---

Nitpick comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Around line 391-401: The download/upload template in documentUploader.jsp
hardcodes the strings "Error" and "Delete" (around file.error and the .delete
button) which bypasses i18n; replace those literals with <fmt:message> lookups
(e.g., keys like "label.error" and "action.delete") in the template so the UI
text is localized, ensuring the message tags are placed where the current span
and button text are rendered (near file.error and the button with class
"delete") and that any attributes or HTML escaping remain unchanged.

In `@src/main/webapp/share/documentUploader/jquery.fileupload-ui.js`:
- Around line 86-90: Replace the deprecated jQuery check $.isArray with the
modern Array.isArray in the code path that inspects data.result.files (the block
returning data.result.files when present); update the condition from
$.isArray(data.result.files) to Array.isArray(data.result.files) and leave the
surrounding logic unchanged so the function still returns data.result.files when
it's an array and [] otherwise.

phc007:
@claude 
Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Around line 302-303: The label's for attribute is pointing to the wrong
control (it uses "destinationDrop") which mis-associates the "Folder" label;
update the label element that currently reads <label for="destinationDrop" ...>
so its for attribute matches the Folder select id "destFolderDrop" (the select
with id/name destFolderDrop and onchange setDestFolder), ensuring the label
targets destFolderDrop for correct focus and accessibility.
- Around line 492-493: Change the loose equality check "data.textStatus ==
'error'" to a strict equality ("data.textStatus === 'error'") and replace the
hardcoded string "Server error" with the i18n message key lookup for
eform.errors.upload.failed (use the same message lookup mechanism used
elsewhere on this page for consistency)

---

Nitpick comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Around line 35-37: Remove the redundant explicit import of ProviderData: the
wildcard import io.github.carlos_emr.carlos.mds.data.* already covers
ProviderData, so delete the line importing
io.github.carlos_emr.carlos.mds.data.ProviderData to avoid duplication; verify
no other duplicate imports remain and keep the remaining imports (including
CommonLabResultData) unchanged.

In `@src/main/webapp/share/documentUploader/jquery.fileupload-ui.js`:
- Around line 186-196: The code incorrectly reads data.result[0].error which
hardcodes the first result for every file; update the error lookup to use the
matching index (data.result[index].error) or remove the redundant override since
file is already derived from files[index] (see variables file, files, index and
the getFilesFromResponse override) so each file gets its corresponding per-file
error even when sequentialUploads is false. Ensure the conditional still guards
for Array.isArray(data.result) and bounds-check that data.result[index] exists
before reading .error.

phc007:
@claude
apply all [penify-dev] comments above

Inline comments:
In `@src/main/webapp/share/documentUploader/jquery.fileupload-ui.js`:
- Around line 554-556: In _editHandler's fail callback the error lookup still
uses data.result[0] which returns the wrong file for multi-file uploads; change
the fallback to use data.result[index] so the line building the error variable
(the var error = ... expression referencing data.files[index] and data.result)
uses data.result[index] instead of data.result[0], ensuring the error
corresponds to the current file being processed.

---

Nitpick comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Around line 431-434: Remove the dead commented-out Cancel button block (the
HTML with class "btn btn-warning cancel", the glyphicon "glyphicon-ban-circle"
and the literal "Cancel" label) since it is unreachable; if you later
reintroduce a Cancel button in documentUploader.jsp, replace the hardcoded
"Cancel" text with the i18n JSTL key (e.g., use <fmt:message
key="global.btnCancel" />) to match the template's localization pattern.

phc007:
@claude 
Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Line 112: The <script> tag currently injects the raw request.getContextPath()
into the src attribute; replace that raw scriptlet with the OWASP encoder by
using Encode.forHtmlAttribute(...) when emitting the context path (i.e. use
Encode.forHtmlAttribute(request.getContextPath()) or
Encode.forHtmlAttribute(context) to build the src) so the src attribute is
OWASP-encoded like the form action does.
- Around line 476-501: The fileuploadalways handler is removing $("#tbodyid
tr:first-child") after every file and may delete the download-template row that
the plugin's done/_transition already inserted; update the handler
(fileuploadalways) to only remove rows that match the upload-template (e.g.,
check a CSS class or data attribute set by the upload template) or verify the
row contains an upload-state indicator before removing, so that rows produced by
the done/_transition are not removed; reference the fileuploadalways handler,
the table body selector "#tbodyid", and the plugin's done/_transition flow when
making the change.

In `@src/main/webapp/share/documentUploader/jquery.fileupload-ui.js`:
- Around line 14-39: The UMD wrapper declares plugin dependencies in the
AMD/CommonJS arrays but the factory function only accepts ($, tmpl), causing
unresolved imports; remove the unused plugin entries from the
define/require dependency arrays inside the IIFE (so only 'jquery' and
'blueimp-tmpl' are declared) 
---

Duplicate comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Around line 360-396: The error message is built from the user-controlled
fileName and reflected into the upload template via file.error (rendered with
the unescaped template token {%=file.error%} in the template-upload); fix by
HTML-encoding the server-side error string when you set it (use
Encode.forHtml(fileName) or wrap the result of MiscUtils.sanitizeFileName(...)
in Encode.forHtml()) so file.error contains an escaped value, or alternatively
change the client template rendering to an escaped token instead of
{%=file.error%}; update the code path that sets file.error (the code referencing
fileName / sanitizeFileName) to use Encode.forHtml before serializing the
response.

---

Nitpick comments:
In `@src/main/webapp/documentManager/documentUploader.jsp`:
- Around line 507-509: The global click handler jQuery('.alert').on('click',
...) hides every .alert on the page; narrow the selector to only the uploader
alerts (referenced as '#msg' and '#msgU' in this file) or to the uploader
container (e.g., a parent like '#uploader') so only those alerts get the
click-to-hide behavior; update the handler to use a scoped selector such as
'#msg, `#msgU`' or '#uploader .alert' and keep the same callback ($(this).hide()).
- Line 93: Remove the unused local variable declaration resourcePath (String
resourcePath = context + "/share/documentUploader/") since it is never
referenced; delete that line and ensure all script/style references continue to
use existing context or ${pageContext.request.contextPath} (do not replace
usages with resourcePath), and run a quick build/grep to confirm no other code
relies on the resourcePath symbol.

In `@src/main/webapp/share/documentUploader/jquery.fileupload-process.js`:
- Line 106: Replace the deprecated jQuery type check in the condition using
$.type(value) with the native typeof operator: locate the condition "if
($.type(value) === 'string' && value.charAt(0) === '@')" in
jquery.fileupload-process.js and change the left-hand check to "typeof value ===
'string'" so the condition reads equivalently but without the deprecated $.type
usage; ensure behavior is otherwise unchanged (still checking value.charAt(0)
=== '@').

In `@src/main/webapp/share/documentUploader/jquery.fileupload-validate.js`:
- Line 86: Replace deprecated jQuery $.type(...) checks with native typeof
checks: locate occurrences of $.type(options.maxNumberOfFiles) === 'number' and
any other $.type(...) usages in this file (e.g., the checks around lines that
validate options like maxNumberOfFiles) and change them to typeof
options.maxNumberOfFiles === 'number' (and analogous typeof checks for other
option properties). Ensure you preserve the original truthiness logic and
operator order when swapping to typeof so behavior remains identical.

In `@src/main/webapp/share/documentUploader/jquery.iframe-transport.js`:
- Around line 86-88: The code using the deprecated jQuery helper should be
updated: replace the use of $.isArray in the paramNames assignment with the
native Array.isArray check so behavior is preserved (i.e., paramNames =
Array.isArray(options.paramName) ? options.paramName : [options.paramName]);
update the expression where paramNames and options.paramName are referenced to
use Array.isArray for the test and ensure no other $.isArray calls remain
(target symbols: paramNames, options.paramName).

yingbull:
@claude does this lgtm

yingbull (approved):

PR #952, opened 2026-04-02
Fix: unable to set a schedule

### LINKED ISSUE #951: [Bug]: rschedule doesn't populate
### Description

rschedule does not populate with the template

### Steps to Reproduce

manual install
demo sql
add a template
enter it in the schedule

### Expected Result

schedule applies

### Actual Result

no schedule applies

### Module/Area

Appointment/Schedule

### Error Messages/Logs

```shell

```

### Version/Branch

develop

### Checklist

- [x] I have searched for similar issues
- [x] I can reproduce this consistently
- [x] I have verified no patient data (PHI) is included in this report
- [x] I am willing to submit a PR to fix this

phc007:
## Description
unable to set a schedule, fix
also i18n and bootstrap and HTML 5

## Related Issues
Fix #951

## Screenshots
carlosScheduleSetting

carlosScheduleTemplateSetting

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Validation now requires selecting at least one day when applying a schedule template.

* **Improvements**
  * Modernized schedule template pages with HTML5, Bootstrap layout, meta updates and improved accessibility.
  * Added localized validation/help texts for English, Spanish, French, Polish and Brazilian Portuguese.
  * Safer client/server encoding, improved live preview, corrected date handling and selection behavior.
  * Multisite UI/authorization tightened and selection behavior refined.

* **Bug Fixes**
  * Deletion and data access blocked for users without required site management access.


phc007:
@claude review and fix to HTML 5 and bootstrap norms

phc007:
@claude apply any remaining bot fixes that may still apply
then apply similar HTML5 Bootstrap formatting and styling treatment for scheduletemplateapplying.jsp
then check why /scheduletemplateapplying.jsp does not actually update the schema.  Specifically a line gets entered into rschedule but day_of_week and avail_hour (and possibly avail_hourB) are left blank and fix
check both files for completeness of i18n tags in all the language files adding tags and values as needed

phc007:
@claude LGTM?

phc007:
@claude Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@src/main/webapp/schedule/scheduletemplatesetting.jsp`:
- Around line 121-135: The code loads all active providers when
grantOnlyCurProviderScheduleData is false, but the TODO leaves out filtering by
site/team so restricted users can see out-of-scope providers; update the else
branch to apply site/team access controls (respecting _site_access_privacy and
_team_access_privacy) by either calling a DAO method that returns only providers
for the current user's allowed sites/teams (e.g., add or use
providerDao.getActiveProvidersBySiteOrTeam(...) ) or by post-filtering the
providerDao.getActiveProviders() result using the current user's allowed
site/team sets (obtainable from session/auth utilities) before assigning to
providers; keep the curProvider path unchanged and ensure
providerDao.getProvider(...) remains used for the single-provider case.

---

Outside diff comments:
In `@src/main/webapp/schedule/scheduletemplateapplying.jsp`:
- Around line 131-173: Before performing any DAO writes in the delete branch,
validate that the current user is authorized to manage the provider indicated by
request.getParameter("provider_no"); i.e., at the top of the if
(request.getParameter("delete") != null && ...) block read provider_no into a
variable and call your existing authorization check (e.g.
userHasAccessToProvider(provider_no) or SiteAccessUtil.isManager(provider_no));
if the check fails, immediately return a 403 or redirect and do not call
rScheduleDao.search_rschedule_current1, rScheduleDao.findByProviderNoAndDates,
rScheduleDao.findByProviderAvailableAndDate,
scheduleDateDao.findByProviderPriorityAndDateRange,
scheduleDateDao.findByProviderAndDateRange, or perform any merges—only proceed
with the DAO operations and final response.sendRedirect after the authorization
check passes.
- Around line 75-77: The JSP declares bMultisites, bgColors, and excludedSites
as page-instance variables (<%! ... %>) which can persist across requests;
change bgColors and excludedSites to request-local variables (e.g., declare
within scriptlet or set as request attributes) and stop using instance scope for
bMultisites; when outputting bgColors in the style attribute (the
"background-color" inline style) wrap values with Encode.forCssString(...) and
when injecting into JavaScript use
Encode.forJavaScript(Encode.forCssString(...)) to ensure proper CSS and JS
encoding (follow the CSS/JS encoding pattern used in ticklerAdd.jsp); update
references that build "style='background-color:...'" and "backgroundColor='...'"
to use these encoder calls.
- Around line 188-195: The displayTemplate function currently injects the
server-side provider_no directly into JS and appends the template name without
URL-encoding, enabling XSS/URL-injection; fix by escaping the server value with
the server-side encoder (wrap request.getParameter("provider_no") with
Encode.forJavaScript(...) where provider_no is inserted into the JS string) and
encode the selected template name on the client before building the query (use
encodeURIComponent(s[s.selectedIndex].value) when constructing the
scheduleDisplayTemplate.jsp?name=... URL in the displayTemplate function).

---

Nitpick comments:
In `@src/main/webapp/schedule/scheduletemplateapplying.jsp`:
- Around line 31-40: This JSP is missing the repo-standard JSP documentation
block; add a JSP comment block immediately after the copyright header and before
the existing directives (before the DOCTYPE and <%@ page ... %> directives)
describing the file purpose, key features, any parameters/attributes (e.g., the
session-scoped bean scheduleRscheduleBean of class
io.github.carlos_emr.RscheduleBean), and include an `@since` tag; ensure the block
follows the project's standard format and appears above the existing tags like
<%@ taglib ... %> and <fmt:setBundle basename="oscarResources"/>.

In `@src/main/webapp/schedule/scheduletemplatesetting.jsp`:
- Around line 31-38: Add a standard JSP documentation comment block immediately
after the file header/copyright and before the top-level directives (i.e.,
before the existing <!DOCTYPE html> and <%@ page ... %> directives and the
<fmt:setBundle> tag). The block should include a short Purpose, Features,
Parameters (if any request or page parameters are used), and an `@since` tag
following the repo-standard format used elsewhere in the project; place it as a
JSP comment block so it does not render to the client.

phc007:
@claude 
Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@src/main/webapp/schedule/scheduletemplateapplying.jsp`:
- Around line 253-255: The i18n messages embedded directly in JavaScript string
literals (e.g., inside function onBtnDelete and other confirm/alert usages) must
be resolved into JSP variables and passed through the OWASP JavaScript encoder
before concatenation into confirm()/alert() calls; for each message (used in
onBtnDelete and the other JS confirm/alert spots) add a <fmt:message ...
var="..."/> and replace the inline tag with the encoded variable via
Encode.forJavaScript (or EL e:forJavaScript) so the string is safe for inclusion
in JS strings.

---

Outside diff comments:
In `@src/main/webapp/schedule/scheduletemplateapplying.jsp`:
- Around line 115-123: The page currently loads provider-scoped data (calls to
SiteDao.getActiveSitesByProviderNo) using request.getParameter("provider_no")
before checking authorization; add an upfront authorization check that validates
the requested provider_no against the current user's allowed providers and
returns a 403 (or forwards to the existing access-denied path) if unauthorized,
before any SiteDao calls and before the bMultisites branch runs. Use the
existing variables isSiteAccessPrivacy and CurProviderNo to determine the
privacy mode and validate provider_no (from request.getParameter("provider_no"))
for non-privileged users; apply the same guard for the other render branch
referenced (the code around the later block that also calls
getActiveSitesByProviderNo at lines 157-177) so no provider-scoped DAO reads
occur without authorization.
- Around line 327-397: The problem: disabled-but-checked site inputs remain
serialized because getSelectAddr() disables checkboxes without unchecking them,
and the week-B path disables the wrong checkbox (it disables week-A like
checksun instead of checksun2). Fix by (1) when disabling a checkbox in
getSelectAddr(), explicitly set .checked = false before .disabled = true (or
alternatively have the serializers skip elements where .disabled is true), and
(2) correct the week-B disable logic to target the week-B checkbox name (use
"check"+s.substring(0,3)+"2" or the equivalent id for addDataStringB()’s addr2
handling) so the proper checkbox is disabled/unchecked. Ensure addDataString()
and addDataStringB() either ignore disabled inputs or see the prior uncheck
change.

---

Duplicate comments:
In `@src/main/webapp/schedule/scheduletemplatesetting.jsp`:
- Around line 138-152: The provider list currently falls back to
providerDao.getActiveProviders() without site/team scoping; replace that with a
site/team-filtered retrieval or post-filter the list so out-of-scope providers
are never shown. Specifically, update the else branch (and the
grantOnlyCurProviderScheduleData branch if needed) to call a scoped DAO method
(e.g.,
providerDao.getActiveProvidersBySite(siteId)/getActiveProvidersByTeam(teamId))
or filter the returned List<Provider> by provider.getSiteId()/getTeamId()
against the current user's site/team (obtain from session/context), ensuring
providers only from the allowed site/team are added to the providers variable
instead of exposing all providers via providerDao.getActiveProviders().

phc007:
@claude Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@src/main/webapp/schedule/scheduletemplateapplying.jsp`:
- Around line 515-525: The current validation only checks
document.schedule.day_of_week and blocks schedules that only set week-B
selections; update the validation in the submit routine (the block that shows
i18n.msgSelectDay) to consider both day_of_week and day_of_weekB (the value
populated by addDataStringB()), i.e., treat the day selection as valid if either
document.schedule.day_of_week.value or document.schedule.day_of_weekB.value is
non-empty; adjust the conditional that triggers the alert so it checks both
fields before returning false.
- Around line 651-653: The localized button labels produced by <fmt:message> are
placed directly into HTML attributes (e.g., the Delete button with onClick
calling onBtnDelete and the other button at the later occurrence) which can
break when translations contain quotes; fix by extracting the messages into JSP
variables (e.g., create var names like deleteLabel and otherButtonLabel using
<fmt:message>), then use Encode.forHtmlAttribute() when inserting those
variables into attribute values (replace direct <fmt:message> usages in the
button value attributes with the encoded variables) so attributes are safe from
embedded quotes.
- Around line 271-273: The URL built in displayTemplate uses JavaScript-escaped
provider_no but not URI-encoded, so special characters can break the query
string; update the construction of url in function displayTemplate to URI-encode
the provider parameter by wrapping the server-inserted provider_no expression
with encodeURIComponent before appending it (same approach used for
templateName) so both name and providerid are properly encoded when composing
the query string.

phc007:
@claude Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@src/main/webapp/schedule/scheduletemplateapplying.jsp`:
- Around line 100-102: bMultisites is declared as a JSP declaration (shared
across requests) but used per-request by getSelectAddr(), causing stale
behavior; remove or stop using the declaration-level bMultisites and instead
compute a local request-scoped boolean in the request scriptlet where
bgColors/excludedSites are created, then pass that local bMultisites value into
the getSelectAddr(...) call (or any helper methods) so the UI/auth path uses the
current request's setting rather than the servlet-instance snapshot.
- Around line 116-136: The request currently skips access checks when
provider_no is missing; before any DAO calls (inside the isSiteAccessPrivacy
branch) validate and reject blank/missing provider_no by checking reqProviderNo
(e.g., if reqProviderNo == null or empty) and call
response.sendError(HttpServletResponse.SC_FORBIDDEN) and return; then reuse that
same reqProviderNo local variable in the subsequent comparisons and calls to
SiteDao.getActiveSitesByProviderNo(CurProviderNo) and
getActiveSitesByProviderNo(reqProviderNo) (ensuring you do not re-read
request.getParameter elsewhere).

yingbull:
@claude does this Lgtm

PR #1006, opened 2026-04-04
chore: remove USE_NEW_ECHART property gate

yingbull:
The new e-chart (Index2.jsp) has been forced on for all users. Remove
the temporary feature gate code (marked "To be removed" in comments)
that conditionally routed between legacy Index.jsp and new Index2.jsp.

- Remove USE_NEW_ECHART and DOCS_NEW_ECHART property reading from Startup
- Simplify EctIncomingEncounter2Action to always return success2
- Remove useNewEchart conditional from Index2.jsp
- Remove dead "success" struts result mapping to legacy Index.jsp
- Remove USE_NEW_ECHART property from both carlos.properties files

## Summary by Sourcery

Remove the legacy e-chart implementation and its feature toggle so the new encounter chart is always used.

Enhancements:
- Always route encounter flows to the new Index2.jsp e-chart instead of conditionally falling back to the legacy page.
- Clean up startup configuration by dropping temporary DOCS_NEW_ECHART and USE_NEW_ECHART property handling and related servlet context attributes.
- Remove obsolete Struts result mappings, save actions, and the legacy encounter Index.jsp view tied to the old e-chart.

Build:
- Stop excluding the legacy encounter Index.jsp from the build configuration so it can be fully removed from the project.

Chores:
- Remove USE_NEW_ECHART configuration keys from carlos.properties files now that the new e-chart is standard.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated the app to a single, modern encounter display interface.
  * Removed legacy encounter view option and its configuration toggle.
  * Simplified encounter page navigation and request handling so users consistently see the updated encounter view.
  * Streamlined startup/session behavior related to choosing the encounter display.


yingbull:
@claude does this lgtm?

PR #1044, opened 2026-04-04
feat: implement embedded WAF for Tomcat 11 with GeoIP, rate limiting, and security

yingbull:
Add a comprehensive 5-layer Web Application Firewall that integrates natively with the Tomcat 11 / Jakarta EE 11 stack:

Layer 1 - Rate Limiting: Tomcat 11 built-in RateLimitFilter (100 req/min global, 10 req/min on login endpoint for brute force protection)

Layer 2 - GeoIP Filtering: MaxMind GeoLite2-Country database restricts access to Canadian IPs only; private/loopback IPs always pass through for Docker/dev envs

Layer 3 - IP Reputation: Spamhaus DROP list loaded into memory with CIDR matching, auto-refreshed every 4 hours via background scheduler

Layer 4 - WAF Request Inspection: OWASP CRS-inspired pattern detection for SQLi, XSS, path traversal, command injection, CRLF header injection, scanner/bot blocking, and request size limits. Healthcare-aware with relaxed paths for clinical notes to avoid false positives on medical terminology.

Layer 5 - Enhanced Security Headers: Adds HSTS (HTTPS-only), CSP (report-only mode), Referrer-Policy (prevents PHI leakage), COOP, CORP to existing ResponseDefaultsFilter. StuckThreadDetectionValve added to Tomcat for DoS protection.

All new components start in detect mode (log only, don't block) for safe production rollout. WAF logs never include parameter values to prevent PHI exposure.


phc007:
@yingbull 
so several things have come up over the years with geo fencing OSCAR 19
1. The (cloud) server you install to may not be in your province / country (hopefully not the US but ya know)
2. Some users require a server in Brazil (Luiz)
3. End users go on vacation and want to check in on their labs (silly doctors)

PR #1046, opened 2026-04-04
feat: add copy-to-clipboard button for patient demographic JSON in EC…

1984Doc:
…hart header

Adds a copy icon (two overlapping rectangles) between the patient name and sex in the EChart header bar. Clicking it copies structured JSON to the clipboard containing patient demographics, logged-in provider info (name, CPSO, OHIP), and clinic details (name, address, phone, fax) for pasting into external referral forms, fax cover sheets, or other systems.

All values are OWASP-encoded server-side via Encode.forJavaScript().

## Checklist

- [ ] My commits are signed off for the [DCO](https://developercertificate.org/) (`git commit -s`)
- [ ] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format, or I've written clear commit messages and will use the format next time
- [ ] I have not included any patient data (PHI) in this PR
- [ ] I have added tests for new functionality, or this change doesn't need new tests
- [ ] I have read the [contributing guide](CONTRIBUTING.md)

## Summary by CodeRabbit

* **New Features**
  * Added a copy-to-clipboard button to the encounter header that allows users to easily copy patient demographic information, provider details, and clinic data.

### DIFF
--- a/src/main/webapp/casemgmt/newEncounterHeader.jsp
+++ b/src/main/webapp/casemgmt/newEncounterHeader.jsp
@@ -40,6 +40,9 @@
 <%@ page import="io.github.carlos_emr.carlos.commn.model.Demographic" %>
 <%@ page import="io.github.carlos_emr.carlos.encounter.pageUtil.EctSessionBean" %>
 <%@ page import="io.github.carlos_emr.carlos.commn.model.Provider" %>
+<%@ page import="io.github.carlos_emr.carlos.commn.model.Clinic" %>
+<%@ page import="io.github.carlos_emr.carlos.commn.dao.ClinicDAO" %>
+<%@ page import="org.owasp.encoder.Encode" %>
 
 <%@ taglib uri="/WEB-INF/security.tld" prefix="security" %>
 <%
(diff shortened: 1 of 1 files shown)


1984Doc:
@claude what bot reviews are valid after you assess them and the actual code, what did they miss that is valid also, and what should be done before merge?

yingbull:
@1984Doc What's the use case on this in this format?

yingbull:
@phc007 @1984Doc where we want to go with this one.

json copy seems unexpected UI wise and pretty niche.
If some block as plain text could be useful, maybe we could do that with a modifier, else this might be so niche a greasemonkey is actually appropriate.

phc007:
@yingbull @1984Doc 
I don't find a usecase for this except as an intermediate step
UX that can't be directly used by the provider (or baffles them) seems a bad idea

1984Doc:
I thought the json would be the easiest option because you're basically
going to write another piece of code for whatever site that you're going to
paste it in and I thought that was the easiest way to move a large amount
of information.

yingbull:
Let's keep hashing this out.  There is a need / interest driving this, so maybe there is wide scope / another way to approach it.

phc007:
The general public will be confused and not amused
If it goes in I would put a code icon on a small button <> and gate it.
Either hidden in carlos.properties, or visible in provider settings would do

yingbull:
Let's hold this for now and look at some way to have power user features like this as we update that screen.

PR #1595, opened 2026-04-12
fix(login): stop storing PIN/nextPage in session; validate providerNo before shared-list mutation

yingbull:
## Summary

Fixes three true-positive trust-boundary findings in `Login2Action.java` flagged by the Semgrep `tainted-session-from-http-request` rule family and SnykCode `java/TrustBoundaryViolation`.

### 1. `CaseMgmtUsers` cross-session leakage (line 665)
The `ArrayList<String>` at `ServletContext` scope is shared across every user session. Login was mutating it in-place with an unvalidated `providerNo` and then storing the same reference into the new user's session — a concurrent login could observe another user's providerNo via the shared list.

- `providerNo` is now regex-validated `[A-Za-z0-9_-]{1,20}` before the shared list is mutated
- List mutation is synchronised on the list
- A defensive copy (`new ArrayList<>(newDocArr)`) is stored in the session so each user's session holds an independent list

### 2. PIN no longer stored in session (line 921)
The forced-password-change flow stashed the user's PIN in session with only a `[0-9]{4}` regex guard, expecting to reuse it after password reset. Sessions can live 30 min; XSS or session fixation in that window would leak a second-factor secret.

- `setUserInfoToSession` no longer writes `pin` to the session at all (both the forced-password-change and MFA callers)
- The forced-password-change return branch now short-circuits after a successful reset: redirects to the login page with a `?msg=password_changed` flag, user re-enters their PIN fresh on the normal login form
- The MFA flow already uses the session-stashed `LoginCheckLogin` object for re-auth, not the raw PIN, so dropping the PIN stash there is also safe

### 3. `nextPage` no longer stored in session (line 930)
A session-stored redirect target is a latent CWE-601 vector — the attribute can be picked up outside the validation context or survive long enough that the originating request is out of scope.

- `setUserInfoToSession` no longer writes `nextPage` to the session
- Redirect to `forcepasswordreset.jsp` carries `nextPage` as a URL query parameter (validated via `RedirectValidationUtils` before inclusion, `Encode.forUriComponent` applied)
- `forcepasswordreset.jsp` re-validates `nextPage` server-side and echoes it as a hidden form field (HTML-attribute-encoded via `Encode.forHtmlAttribute`)
- On return to `login.do`, `nextPage` is read from the request parameter and re-validated one more time before use

## Changed

- `src/main/java/io/github/carlos_emr/carlos/login/Login2Action.java`:
  - `setUserInfoToSession` signature: 4 args → 2 args (dropped `pin` and `nextPage`)
  - Both callers (forced-password-change at line 561, MFA at line 621) updated
  - Forced-password-change return branch rewired to redirect to login page after reset instead of falling through to `cl.auth`
  - Redirect to `forcepasswordreset.jsp` now includes validated `nextPage` as query param
  - `removeAttributesFromSession` keeps clearing `pin`/`nextPage` keys for backward-compat with sessions created before this change
  - `CaseMgmtUsers` handling validated + defensive-copied

- `src/main/webapp/forcepasswordreset.jsp`:
  - Reads `nextPage` from request parameter, re-validates via `RedirectValidationUtils`
  - Echoes validated value as hidden form field (HTML-attribute-encoded)

## Why not just suppress?

All three were classified as **true positives** in the security-audit review. They're not scanner noise — each one represents a real (if small-window) data-exposure risk. Suppressing would have hidden the signal.

## Test plan

- [ ] `make install --run-unit-tests` green
- [ ] Manual: trigger forced-password-change flow, confirm user is redirected to login page after reset and must re-enter credentials
- [ ] Manual: trigger forced-password-change with `nextPage=/somewhere` query param, confirm it survives the round-trip and redirects correctly after re-login
- [ ] Manual: trigger MFA flow, confirm it still works without the PIN in session
- [ ] Manual: confirm existing session attrs from before this change are cleared by `removeAttributesFromSession` (backward-compat)
- [ ] Re-scan: confirm Semgrep/Snyk trust-boundary alerts at lines 665, 921, 930 clear

## UX note

Users in the forced-password-change flow will now be prompted to log in again after resetting their password (rather than being transparently re-authenticated). This is intentional — the re-login path ensures PIN re-entry happens through the normal authenticated channel rather than via a session-stashed copy. Forced-password-change is a rare event (initial-account setup), so the extra step is acceptable.

---
## Summary by cubic
Stops storing PIN and nextPage in session and tightens `providerNo` and redirect handling to close trust-boundary gaps in the login flow. Adds a safe post-reset re-login via postLoginRedirect and shows a clear success message.

- **Bug Fixes**
  - Removed PIN from session; after password reset, redirect to login (`?msg=password_changed`) and require fresh PIN. Carry the original destination as validated `postLoginRedirect` through the login form and redirect to it after successful login for single/no-facility users; suppressed the CodeQL redirect alert on this validated path.
  - Stopped storing `nextPage` in session; validate it, pass via query param to `forcepasswordreset.jsp`, re-validate in the JSP, and echo as a hidden field; preserve it on validation-error redirects back to the reset page.
  - Validated `providerNo` with `[A-Za-z0-9_-]{1,20}` before mutating the shared `CaseMgmtUsers` list; handle null list, synchronize mutation, and store a defensive copy in session inside the synchronized block.
  - Simplified `setUserInfoToSession` to `(userName, password)`; password is hashed; legacy `pin`/`nextPage` keys still cleared for backward compatibility.
  - Show password-changed success banner on `index.jsp`; added i18n key `loginApplication.passwordChangedLabel` in `oscarResources_*.properties`.

<sup>Written for commit 40c7d33b4f06931d9acb678dfe94e41959af090e. Summary will update on new commits.</sup>

### DIFF
--- a/src/main/java/io/github/carlos_emr/carlos/login/Login2Action.java
+++ b/src/main/java/io/github/carlos_emr/carlos/login/Login2Action.java
@@ -360,6 +360,12 @@ public final class Login2Action extends ActionSupport {
         if (request.getParameter("forcedpasswordchange") != null
                 && request.getParameter("forcedpasswordchange").equalsIgnoreCase("true")) {
             // Coming back from force password change.
+            //
+            // Only userName and (hashed) password are retrieved from session; pin and
+            // nextPage are NOT stored in session (see setUserInfoToSession Javadoc).
+            // nextPage is propagated via the hidden form field carried by
+            // forcepasswordreset.jsp; pin is re-entered by the user after a successful
+            // reset via a fresh login on the login page.
             userName = (String) request.getSession().getAttribute("userName");
 
             // Username is only letters and numbers
@@ -369,17 +375,12 @@ public final class Login2Action extends ActionSupport {
 
             password = (String) request.getSession().getAttribute("password");
 
-            pin = (String) request.getSession().getAttribute("pin");
-
-            // pins are integers only
-            if (!Pattern.matches("[0-9]{4}", pin)) {
-                pin = "";
-            }
-            nextPage = (String) request.getSession().getAttribute("nextPage");
-            // Validate nextPage retrieved from session to prevent open redirect (CWE-601 defense in depth)
+            // nextPage is a request parameter (hidden form field), not a session attribute
+            nextPage = request.getParameter("nextPage");
+            // Validate nextPage to prevent open redirect (CWE-601)
             if (!RedirectValidationUtils.isValidRelativeRedirect(nextPage)) {
                 if (nextPage != null) {
-                    logger.warn("Rejected invalid nextPage from session: {}", LogSanitizer.sanitize(nextPage));
+                    logger.warn("Rejected invalid nextPage on password-reset return: {}", LogSanitizer.sanitize(nextPage));
                 }
                 nextPage = null;
             }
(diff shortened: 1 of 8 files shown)


yingbull:
@claude what review comments are valid, why or why not, what action to take and why, and after your own review what still needs to be fixed or done that was not identified.   give a numbered summary list grouped into must, should, could do.

yingbull:
@claude fix the should

yingbull:
@copilot+claude-opus-4.7 examine current develop branch did our migration of struts config and to web-inf dots address this, and if not, provide a clear and complete plan on what would need to be done to resolve this based on that up to date branch.

yingbull:
will revisit in separate pr.

PR #1996, opened 2026-04-26
Fix vaccine brands JSON 404

### LINKED ISSUE #1991: Static .json files served via JSP servlet 404 — affects vaccine-brands.json + DataTables i18n
## Problem

`web.xml` maps `*.json` to the JSP servlet:

```xml
<servlet-mapping>
    <servlet-name>jsp</servlet-name>
    <url-pattern>*.jsp</url-pattern>
    <url-pattern>*.jspf</url-pattern>
    <url-pattern>*.json</url-pattern>
</servlet-mapping>
```

This is intentional — CARLOS has many `.json` files that **are** JSPs that emit JSON responses (e.g. `/admin/clinicNbrManage.json`, `/admin/api/clientManage.json`, `/provider/tokenManage.json`, `/admin/keygen/getPublicKey.json`). Those work fine.

The problem is **static `.json` data files** under `webapp/`. When requested via HTTP, they get routed to Jasper, which tries to compile them as JSPs, fails (because they start with `[` or `{`), and Tomcat returns 404. The static file is never served.

## Affected files

After this PR, only one remaining static `.json` file in the webapp is broken:

- `src/main/webapp/prevention/vaccine-brands.json` (12 KB) — the bundled vaccine catalogue used by `prevention/index.jsp`. Currently returns **404** when fetched at `/carlos/prevention/vaccine-brands.json`. The JSP at `prevention/index.jsp:613` falls back to this URL when the admin's custom upload via `eform/displayImage?imagefile=vaccine-brands.json` is unavailable, so the failure is silent for sites with a custom upload, but breaks the bundled catalogue for sites without one.

The 5 DataTables i18n JSON files were affected by the same bug and are fixed on `chore/billing-refactor` (commit `59e4fa2824`):
- `src/main/webapp/library/DataTables/i18n/en-GB.json`
- `src/main/webapp/library/DataTables/i18n/es-ES.json`
- `src/main/webapp/library/DataTables/i18n/fr-FR.json`
- `src/main/webapp/library/DataTables/i18n/pl.json`
- `src/main/webapp/library/DataTables/i18n/pt-BR.json`

## Reproducing

```bash
curl -i http://localhost:8080/carlos/prevention/vaccine-brands.json
# HTTP/1.1 404
# Content-Type: text/html;charset=UTF-8
```

## Fix

Two viable approaches; the proven one (used for the DataTables i18n files):

**Option 1 — minimal: prepend a `contentType` JSP directive (one-line change per file)**

Add this single line at the top of each affected static `.json` file:

```jsp
<%@ page contentType="application/json; charset=UTF-8" %>
```

Jasper compiles it as a trivial JSP that emits the JSON literal verbatim with the correct `Content-Type` header. No `web.xml` / `struts.xml` / JSP-side URL changes required.

This is the approach used in commit `59e4fa2824` for the DataTables i18n files and verified end-to-end via Playwright (DataTables loads correctly with localized labels).

**Option 2 — broader: route static `.json` paths to the default servlet**

A path-prefix mapping like `<servlet-mapping><servlet-name>default</servlet-name><url-pattern>/library/*</url-pattern></servlet-mapping>` would route static-asset paths to Tomcat's default servlet. Per Jakarta Servlet 6 precedence rules path > extension, so this overrides the `*.json` extension routing. **Not recommended without care**: in this PR an attempt at this approach overrode Tomcat's implicit `/` mapping for the default servlet and broke the home page. Would need a fully-specified `default` servlet declaration to be safe.

## Recommended action

Apply Option 1 to `vaccine-brands.json`:

```bash
{ echo '<%@ page contentType="application/json; charset=UTF-8" %>'; cat src/main/webapp/prevention/vaccine-brands.json; } > /tmp/vb.json && mv /tmp/vb.json src/main/webapp/prevention/vaccine-brands.json
```

Verify:
```bash
curl -i http://localhost:8080/carlos/prevention/vaccine-brands.json
# Should return 200 with Content-Type: application/json
```

## Prevention

A one-line lint check could prevent regressions: any `webapp/**/*.json` file whose first non-blank line doesn't start with `<%` should be flagged. Could live alongside `scripts/lint/check-jsp-size.sh` or `check-encoder-null-safety.sh`.

HassanMahra:
Fixes #1991.

This keeps vaccine-brands.json available through the existing *.json JSP mapping, with the JSON content type expected by the prevention page. It also adds the _prevention read check requested in review and trims directive whitespace so the response starts with the JSON array.

Validation:
- git diff --check
- mvn -DskipTests=true -T 1C package war:exploded -Pjspc

I left the broader JSON lint/Biome guard out of this small fix because existing webapp/**/*.json JSP endpoints use a few different directive patterns. That cleanup is better as a separate PR.

---
## Summary by cubic
Fixes the 404 for vaccine-brands.json and serves it via the existing *.json JSP mapping with the correct JSON content type. Adds the `_prevention` read check (returns 403 when missing) and trims JSP directive whitespace so the response starts with the JSON array.

<sup>Written for commit cc7bc490c3a905a096d6168e71bb1668851fb84e. Summary will update on new commits.</sup>

### DIFF
--- a/src/main/webapp/prevention/vaccine-brands.json
+++ b/src/main/webapp/prevention/vaccine-brands.json
@@ -1,6 +1,8 @@
 <%@ page contentType="application/json; charset=UTF-8" trimDirectiveWhitespaces="true" %><%
-    if (!io.github.carlos_emr.carlos.utility.SpringUtils.getBean(io.github.carlos_emr.carlos.managers.SecurityInfoManager.class)
-            .hasPrivilege(io.github.carlos_emr.carlos.utility.LoggedInInfo.getLoggedInInfoFromSession(request), "_prevention", "r", null)) {
+    io.github.carlos_emr.carlos.utility.LoggedInInfo loggedInInfo =
+            io.github.carlos_emr.carlos.utility.LoggedInInfo.getLoggedInInfoFromSession(request);
+    if (loggedInInfo == null || !io.github.carlos_emr.carlos.utility.SpringUtils.getBean(io.github.carlos_emr.carlos.managers.SecurityInfoManager.class)
+            .hasPrivilege(loggedInInfo, "_prevention", "r", null)) {
         response.sendError(403);
         return;
     }
--- a/src/main/webapp/prevention/vaccine-brands.json
+++ b/src/main/webapp/prevention/vaccine-brands.json
@@ -1,3 +1,4 @@
+<%@ page contentType="application/json; charset=UTF-8" %>
 [
 {"name":"OtherA","value":"IXCHIQ 3 log10 TCID50 per 0.5ml powder for solution","manufacture":"Valneva Austria GmbH","dose":"0.5","units":"mL","route":"IM","din":"2548984"},
 {"name":"Chol-Ecol-O","value":"DUKORAL oral","manufacture":"Valneva Sweden AB","dose":"","units":"DOSE","route":"PO","din":"2247208"},
--- a/src/main/webapp/prevention/vaccine-brands.json
+++ b/src/main/webapp/prevention/vaccine-brands.json
@@ -1,5 +1,10 @@
-<%@ page contentType="application/json; charset=UTF-8" %>
-[
+<%@ page contentType="application/json; charset=UTF-8" trimDirectiveWhitespaces="true" %><%
+    if (!io.github.carlos_emr.carlos.utility.SpringUtils.getBean(io.github.carlos_emr.carlos.managers.SecurityInfoManager.class)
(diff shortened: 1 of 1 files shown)


phc007:
@yingbull @HassanMahra 
This may not be maintainable.  You are suggesting that all JSON files be turned into jsp's who then get stripped for content
 
The JSON extension should reflect JSON content and be served as a static asset 
There are 4 files with a .json extension that provide JSON *output* rather than content and should have a jsp extension, and we propose renaming those x.json.jsp
This is consistent with existing conventions the same way that that x.js.jsp naming indicates servlet augmented javascript

see [pull #2029](https://github.com/carlos-emr/carlos/pull/2029)

PR #2335, opened 2026-05-27
fix: redirect to login page after logout instead of server-side forward

### LINKED ISSUE #2245: [Bug]: logout redirects to logout causing flashing content
### Description

logout notification should always redirect to the login and stay there

### Steps to Reproduce

allow timeout logout
this doesn't close the window but if you refreash or navigate 
you get a logout that redirects to logout which is too fast to actually interact with

### Expected Result

logout to login page and stay there

### Actual Result

logout > logout > logout loop

### Module/Area

Other

### Error Messages/Logs

```shell
none but there is a 302 redirection of GET logout to GET logoutPage ?
```

### Version/Branch

_No response_

### Checklist

- [x] I have searched for similar issues
- [x] I can reproduce this consistently
- [x] I have verified no patient data (PHI) is included in this report
- [ ] I am willing to submit a PR to fix this

CecilFoubert:
## Summary

- Replaces the server-side Struts forward (`/WEB-INF/jsp/login/index.jsp`) on the `logout` action with an HTTP redirect to `/index`
- The old forward kept the browser URL at `/logout`, and because `logout.jsp` (served by `/logoutPage`) JS-redirects the browser back to `logout` after 500 ms, users experienced a redirect loop that flashed too fast to interact with
- The new redirect issues a proper 302 to `/index`, which is exempt from `LoginFilter` session checks, renders the login page at the correct URL, and breaks the loop
- Updates the stale JavaDoc example in `Logout2Action` (previously showed `/logoutPage` as the result, which was the original source of the loop)

## Test plan

- [ ] Log out normally: confirm browser lands on `/index` (login form) with correct URL
- [ ] Let session expire, navigate to a page, confirm: `logoutPage` → `logout.jsp` 500 ms → `/logout` → 302 → `/index` (login form, no loop)
- [ ] Run the Playwright login/reset regression script (`npm run test:login-playwright`) against a local dev environment
- [ ] Run unit tests: `mvn test -Dgroups="unit"`

Fixes #2245

## Summary by Sourcery

Enforce logout as a POST-only mutating action and redirect users to the login page at /index after logout to avoid redirect loops and ensure correct session cleanup.

Bug Fixes:
- Prevent redirect loops by changing the logout Struts result from a server-side forward of the login JSP to an HTTP redirect to /index.
- Ensure logout side effects only occur on POST by rejecting non-POST requests with 405 in Logout2Action and wiring logout.jsp to submit a hidden POST form instead of using a GET redirect.

Enhancements:
- Document the logout behavior and POST-only guard more clearly in Logout2Action JavaDoc and logout.jsp comments.
- Register Logout2Action in the mutator contract test suite to verify it rejects unsafe GET access as a mutating endpoint.

Tests:
- Extend the mutator action contract test to include Logout2Action as an unconditional mutator endpoint.

---
## Summary by cubic
Redirects logout to `/index` with a 302 and makes `/logout` POST-only to stop the redirect loop and ensure correct session cleanup. Users now land on the login page at the right URL, and the `LoginFilter` exemption applies.

- **Bug Fixes**
  - Enforce POST on `/logout` (405 on non-POST with `Allow: POST`; uses `Locale.ROOT` for method check); update JavaDoc and add `@Override` on `execute()`; register `io.github.carlos_emr.carlos.login.Logout2Action` in `MutatorActionGetRejectionContractTest`.
  - Change Struts `logout` result to `type="redirect"` → `/index`; update `logout.jsp` to POST via a hidden form with an explicit context path and remove meta-refresh; exclude `/logout` from CSRFGuard.

<sup>Written for commit 2f8ab58625452f9dae063673f799b6edad521d97. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/2335?utm_source=github">Review in cubic</a>


CecilFoubert:
@coderabbitai Both major findings were addressed in the second commit ([`0cd6266`](https://github.com/carlos-emr/carlos/commit/0cd6266d10)) pushed after your review. Addressing each:

**"Add GET/HEAD rejection" (lines 184–218)** — Done. `execute()` now rejects any non-POST request with HTTP 405 as the very first thing it does, before `session.invalidate()` or cookie deletion can fire:

```java
if (!"POST".equalsIgnoreCase(request.getMethod())) {
    response.sendError(HttpServletResponse.SC_METHOD_NOT_ALLOWED, "POST required");
    return NONE;
}
```

`Logout2Action` is also registered in `unconditionalMutators()` in `MutatorActionGetRejectionContractTest` with an explanatory comment about the empty privilege tuple (since `hasPrivilege()` is never called here, the contract assertion skips the privilege check).

**"`hasPrivilege()` omission" (lines 111–123)** — Intentionally absent; this is an approved exception to the standard rule, documented in the JavaDoc on `execute()`. `/logout` is listed in `LoginFilter.EXEMPT_URLS` specifically because it must handle requests after the session has already expired (session-timeout flow). Calling `hasPrivilege()` on a null/expired `LoggedInInfo` would throw or block the very cleanup this action exists to perform. The inline JavaDoc explains the reasoning so future maintainers don't re-add the check reflexively.

CecilFoubert:
@sourcery-ai The duplication of `/index` across the Struts XML and the JavaDoc is intentional and can't be practically centralized. The Struts XML is declarative config — there's no mechanism for a Java constant to be referenced from within an XML result attribute. The JavaDoc occurrence is a documentation example of that config, not a second live reference. A Java constant would only cover the Java side while the XML entry would still hardcode the path anyway, so you'd still have two places to update on a path change. If the login path ever moves, `grep -r '/index' src/main/webapp/WEB-INF/classes/struts-login.xml` catches it immediately.

CecilFoubert:
@github-code-quality Fixed in commit f3a826d — `@Override` added directly above `execute()`.

CecilFoubert:
@github-advanced-security Fixed in commit 19badd6 — replaced `equalsIgnoreCase("POST")` with `"POST".equals(request.getMethod().toUpperCase(Locale.ROOT))`. The servlet container always returns uppercase ASCII for HTTP methods, but the explicit `Locale.ROOT` normalization makes this immune to locale-sensitive Unicode case mapping (e.g. Turkish dotted-I) and satisfies the `IMPROPER_UNICODE` rule.

CecilFoubert:
I expect the problem may have come from using a timer for the logout AND having a refresh-redirect to logout page. Basically, when you logout, it would render the login page, but you'd still be at the /logout URL, and the two mechanisms for logging out would race to try and log you out. So the timer in the jsp would go, time you out, and refresh you, which would trigger the refresh-redirect, which would activate the timer, and repeat ad nauseum. Because the jsp timer is 500ms, it made a flashing effect. Fixed by swapping the logout redirect to a proper HTTP 302 code, which sends it to the login page Properly (without the /logout), so it doesn't start looping. : )

CecilFoubert:
@github-advanced-security Alert #25697 (line 124) is the same `IMPROPER_UNICODE` finding as #25695, re-raised on commit `f3a826d` after the line shifted by one when `@Override` was added. Both were resolved in commit `19badd6` — `equalsIgnoreCase("POST")` replaced with `"POST".equals(request.getMethod().toUpperCase(Locale.ROOT))`.

CecilFoubert:
@sourcery-ai All three points addressed in commit 2f8ab58:

1. **Inaccurate no-JS comment (the flagged issue)** — Fixed. The old text said "Without JavaScript the broadcast still fires via BroadcastChannel (if supported)" which is wrong — BroadcastChannel is JavaScript and can't fire without it. Corrected to: "without JavaScript neither the BroadcastChannel broadcast nor the form POST occurs; the session will expire naturally."

2. **`Allow: POST` header on 405** — Added `response.setHeader("Allow", "POST")` before `sendError`. RFC 7231 requires this on 405 responses.

3. **Form action context path** — Changed `action="logout"` to `action="${pageContext.request.contextPath}/logout"` for explicitness and consistency with the CARLOS pattern used elsewhere (e.g. `mainMenu.jsp`).

4. **Stray `<meta http-equiv="refresh">` also removed** — The HTML head still had the old meta-refresh tag that was supposed to have been removed in the previous commit. It was inconsistent with the comment above it ("There is no meta-refresh fallback") and would have issued a GET to `/logout` after 1s, hitting the 405 guard. Now fully removed.

yingbull (approved):

yingbull:
@CecilFoubert need DCO signoff, you can do it manually id needed on the PR with the statement but commit with -s is best.

yingbull:
@copilot+gpt-5.4 signoff DCO

CecilFoubert:
ah ok

yingbull (approved):

PR #2763, opened 2026-06-10
fix: unguarded Integer.parseInt(demographic_no)

### LINKED ISSUE #2757: [Bug] AddPrevention2Action.validate() — unguarded Integer.parseInt(demographic_no) throws NumberFormatException if parameter is null or non-numeric (claude assist)
## Summary

`AddPrevention2Action.validate()` calls `Integer.parseInt(demographic_no)` without first checking whether `demographic_no` is null or numeric. If the request omits the parameter (or passes a non-numeric string), `Integer.parseInt(null)` throws `NumberFormatException` and the method exits with a 500 Internal Server Error instead of adding a validation error to the result list.

**File:** `src/main/java/io/github/carlos_emr/carlos/prevention/pageUtil/AddPrevention2Action.java`
**Line:** 314

```java
if (!demographicDao.clientExists(Integer.parseInt(demographic_no))) {
    result.add("Patient not found");
}
```

## Type / Priority
- **Type:** Bug
- **Priority:** 3-Medium
- **Kind:** Task

## Why This Is a Problem

`request.getParameter("demographic_no")` returns `null` when the parameter is absent from the request. `Integer.parseInt(null)` immediately throws `NumberFormatException`. Because `validate()` is not wrapped in an exception handler, this propagates as an unhandled 500 rather than a user-visible validation message.

Compare with the correct pattern in `RtlPreventions2Action` (lines 109–120), which validates the parameter with a regex check and a try/catch before any `parseInt` call.

## Steps to Reproduce

1. POST to `prevention/AddPrevention` omitting the `demographic_no` parameter (e.g., via a direct curl/Burp request).
2. Observe: HTTP 500 instead of a validation error response.

## Suggested Fix

Add a null/format check before the `parseInt` call, consistent with the validation style already used in the same `validate()` body:

```java
if (demographic_no == null || !demographic_no.matches("\\d+")) {
    result.add("Invalid or missing demographic_no");
} else if (!demographicDao.clientExists(Integer.parseInt(demographic_no))) {
    result.add("Patient not found");
}
```

This also prevents a second `NumberFormatException` at lines 264–265 (`consentDao.findByDemographic...`) and 275 (`DHIR.getFhirBundleBuilder`) which also call `Integer.parseInt(demographic_no)` unconditionally.

## Related Issues
- #2595 — Similar unguarded Integer.parseInt on layout parameters in EctSaveEncounter2Action
- #2648 — Similar unguarded Integer.parseInt on scheduletemplatecodesetting.jsp

snabxo:
Closes #2757 

## Description

Applied the suggested fix to the unguarded Integer.parseInt(demographic_no) bug.

## Checklist

## Summary by Sourcery

Bug Fixes:
- Prevent NumberFormatException by validating demographic_no is present and numeric before parsing and checking if the patient exists.

---
## Summary by cubic
Validate demographic_no before parsing to prevent crashes and give a clear error message. Null or non-numeric values now add "Invalid or missing demographic_no"; otherwise we check if the patient exists.

<sup>Written for commit f206a531b0c38e27b633763bdee662d4b04aed83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/2763?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/main/java/io/github/carlos_emr/carlos/prevention/pageUtil/AddPrevention2Action.java
+++ b/src/main/java/io/github/carlos_emr/carlos/prevention/pageUtil/AddPrevention2Action.java
@@ -20,7 +20,7 @@
  * McMaster University
  * Hamilton
  * Ontario, Canada
- 
+
  * <p>
  * Now maintained by the CARLOS EMR Project (2026+).
  * https://github.com/carlos-emr/carlos
@@ -311,7 +311,9 @@ public class AddPrevention2Action extends ActionSupport {
         }
 
         DemographicDao demographicDao = SpringUtils.getBean(DemographicDao.class);
-        if (!demographicDao.clientExists(Integer.parseInt(demographic_no))) {
+        if (demographic_no == null || !demographic_no.matches("\\d+")) {
+            result.add("Invalid or missing demographic_no");
+        } else if (!demographicDao.clientExists(Integer.parseInt(demographic_no))) {
             result.add("Patient not found");
         }
 


Ben-Heerema:
@snabxo Thanks for taking this on we appreciate all the help we can get. The current patch is headed in the right direction: it prevents the missing/non-numeric demographic_no cases from reaching Integer.parseInt().

One remaining gap: demographic_no.matches("\\d+") still allows numeric strings outside the int range, so something like 99999999999 can still throw NumberFormatException. Since the issue points to RtlPreventions2Action as the model, I think this should follow the same pattern: validate the shape, parse inside a try/catch, add "Invalid or missing demographic_no" on parse failure, and reuse the parsed int for clientExists().

It would also be good to add a focused regression test for:

- missing demographic_no
- non-numeric demographic_no
- numeric overflow

Thanks again for your interest in our project :)

alphin-08:
Hi, I would like to take this issue if it is still available. I have worked on a few similar bugs in this codebase already and feel comfortable tackling this one.

Issue #2814, opened 2026-06-11
[Security] Missing securityInfoManager.hasPrivilege() across JAX-WS SOAP services — most service methods lack RBAC checks (claude assist)


Ben-Heerema:
## Summary

CARLOS's SOAP services (`/ws/*`) authenticate callers via WSS4J `UsernameToken` interceptors, but the majority of individual service methods do not call `securityInfoManager.hasPrivilege()` before executing business logic. Any valid CARLOS credential holder can invoke any SOAP operation regardless of their assigned role.

## Affected Files

All JAX-WS services under `src/main/java/io/github/carlos_emr/carlos/webserv/ws/`:

- `AppointmentWs` — appointment read/write operations
- `DemographicWs` — full patient demographic access (including `getAllDemographics()` — see #2804)
- `DrugWs` — prescription and drug record operations
- `FacilityWs` — facility administration
- `LabUploadWs` — lab result ingestion
- `LoginWs` — authentication service
- `MeasurementWs` — clinical measurement read/write
- `PrescriptionWs` — prescription management
- `ProgramWs` — program/case management
- `ProviderWs` — provider profile management
- `ScheduleWs` — appointment scheduling
- `SecurityWs` — security object listing
- `DocumentWs` — document management

Exceptions already filed: `AllergyWs` (#2471), `PreventionWs` (#2761).

## Details

The base class `AbstractWs` provides `getLoggedInInfo()` via `LoggedInInfo.getLoggedInInfoFromRequest()`, establishing authenticated identity. However, `AbstractWs` contains no built-in privilege enforcement, and individual service methods overwhelmingly omit the required:

```java
securityInfoManager.hasPrivilege(loggedInInfo, "_secobj", "r/w", null)
```

This is the SOAP equivalent of the REST finding in #2798.

## Impact

2-High — A CARLOS user with any role (receptionist, billing clerk, read-only OAuth consumer) can:
- Read and write prescriptions, measurements, and lab results via `DrugWs`, `MeasurementWs`
- Read all patient demographics via `DemographicWs`
- Manage appointments and schedules via `AppointmentWs`, `ScheduleWs`
- Access and modify program/case management data via `ProgramWs`

WSS4J only confirms the caller is a valid user. Role-based access control requires `hasPrivilege()` at the service-method level.

## Fix

[...]

PR #2981, opened 2026-06-22
[codex] Harden SOAP RBAC checks

CecilFoubert:
## What changed

Adds explicit SOAP service privilege checks through a shared `AbstractWs.requirePrivilege(...)` helper and applies them to the JAX-WS endpoints for demographics, documents, prescriptions, measurements, appointments, lab upload, PM/program reads, provider properties, and the facility listing.

Guards are **patient-scoped wherever a patient is knowable**, not just object-level — see below for why that distinction is the whole point.

Full per-endpoint privilege map, the two security objects deliberately *not* used, and the operator upgrade checklist: **[`docs/soap-rbac-hardening.md`](docs/soap-rbac-hardening.md)**.

## Why

Issue #2814 identified that authenticated SOAP callers were reaching service methods without explicit role-based access checks at the service boundary. WSS4J authentication confirms identity, but these endpoints still need `securityInfoManager.hasPrivilege(...)` enforcement.

## Root cause

The SOAP base class exposed authenticated `LoggedInInfo`, but most individual service methods did not enforce privilege checks before delegating into business logic.

## Object-level is not enough: patient scope

`hasPrivilege` consults a `_object$<demographicNo>` override row **only when a demographic scope is passed**. An endpoint that passes null therefore lets a caller holding the general privilege reach patients a per-patient NORIGHTS row was meant to hide. Guards are scoped three ways:

- **From the request** where the method takes a `demographicId`.
- **From the loaded record** where the patient is only knowable after the load (`getPrescription`, `getMeasurement`, `getDocument`) — re-checked before any transfer is built.
- **Per record** on search and bulk-sync endpoints.

Search and bulk endpoints **filter** rather than throw: failing on the first restricted match would confirm that the restricted patient matched the window, which is exactly what the override exists to hide. Endpoints where the caller named a patient throw, because a short answer would misrepresent the record.

## Policy decisions in this PR

Included:

- `_demographic` for `DemographicWs` — scoped ID reads, per-patient filtering on searches **and on both bulk active exports**
- `_edoc` for `DocumentWs` — scoped where the patient is a parameter, re-checked from the `ctl_document` link row on `getDocument`, filtered on bulk sync
- `_rx` for `PrescriptionWs` — unscoped check at entry, demographic-scoped re-check after load, bulk sync filtered
- `_measurement` for `MeasurementWs` — scoped reads and writes, loaded-record re-check on `getMeasurement`, bulk sync filtered
- `_appointment` for `BookingWs` and `ScheduleWs`
- `_lab w` for all nine `LabUploadWs` upload methods
- `_pmm.programList` / `_pmm.staffList` for `ProgramWs`
- `ProviderWs.getLoggedInProviderTransfer()` by `_pref r`
- `ProviderWs.getProviderProperties(...)`: self reads `_pref r`, cross-provider reads `_admin r`
- `FacilityWs.getAllFacilities(...)` by `_admin r`

Deferred, because policy remained unclear after review:

- `ProviderWs.getProviders2(...)` — admin-only, or readable by any authenticated provider?
- `FacilityWs.getDefaultFacility()` **and its deprecated `getDefaultFacilities()` alias** — admin-only, or tied to authenticated facility membership?

## Security objects deliberately not used

Two candidate objects are granted to **no role in any migration**, so gating on them denies every caller including admin. `SecurityObjectSeedContractUnitTest` lists both under `KNOWN_UNREACHABLE`.

- **`_appointment.UpdatedAfterDate`** is a consent-*bypass* marker, not an access gate: in `ScheduleManagerImpl.getAppointmentUpdatedAfterDate` its absence means "filter by patient consent" and its presence means "you may see results unfiltered". `getAppointmentArchivesUpdatedAfterDate` uses `_appointment r`, matching its non-archive sibling.
- **`_pmm_management`** is referenced by several PMmodule actions but granted by no migration. `ProgramWs` uses `_pmm.programList` / `_pmm.staffList` — the objects the PMmodule UI uses for the same two screens, both granted to `admin` and `doctor` in Ontario and BC.

## Upgrade impact

These endpoints were previously reachable by any authenticated SOAP caller, so some integrations will start faulting. `docs/soap-rbac-hardening.md` carries the full checklist; the two that need action before deploying:

1. **Lab intake** — `_lab` is granted to the `doctor` role only. Dedicated lab-uploader service accounts on minimal custom roles will fault, and from the EMR side intake simply stops; the failure is visible only to the sender. Grant `_lab w` to those roles.
2. **Facility directory / cross-provider property reads** — `_admin` is granted to the `admin` role only. Integrator / inter-EMR sync accounts on provider-type roles will fault on `getAllFacilities` and on cross-provider `getProviderProperties`.

Denials are now audited as `ws.accessDenied` with object, privilege and patient scope (identifiers only), so after deploying, each integration needing a grant appears in the audit log by name rather than failing silently.

## Validation

- `mvn test -Dtest=AbstractWsPrivilegeUnitTest` — **31 tests, 0 failures**.

  The test class was previously named `AbstractWsPrivilegeTest`, which matched none of the surefire `<includes>` in `pom.xml` — CI ran **zero** of its cases and the green check did not cover this file. Renamed to match `**/*UnitTest.java`.

  Coverage includes deny paths for `BookingWs.bookAppointment`, `LabUploadWs.uploadCLS`, `DocumentWs.getDocument`, both `ProgramWs` methods, `FacilityWs.getAllFacilities`, the `getDemographics` batch read, the demographic search gate, the missing-prescription case, the real `SpringUtils`-backed manager resolution, and the per-patient filtering on every bulk endpoint.
- Full unit suite green.
- `scripts/lint/check-bdd-test-naming.sh` and `scripts/lint/check-security-exception-message-convention.sh` — both pass.

An earlier revision of this description cited `LabUploadWsRegressionTest` as validating this change. It does not: it predates this PR (#2288) and invokes the private `importLab` through `ReflectionTestUtils.invokeMethod`, bypassing the new public-method guards entirely. It is run as a regression check only.

## Follow-up

Raised during review, deliberately out of scope here:

- `SecurityInfoManagerImpl.hasPrivilege` swallows infrastructure exceptions and returns `false`, so "check errored" is indistinguishable from "denied". In the per-patient filters that means a transient DB failure silently drops a record. Fixing it changes fail-closed behaviour application-wide.
- Per-record filtering costs one uncached `hasPrivilege` call per record, and `filterReadableDocuments` additionally repeats a `ctl_document` lookup that `DocumentTransfer.getTransfers` also performs. If that cost bites, the fix is a cheaper override lookup in `SecurityInfoManager`, not dropping the filter.
- A migration seeding `_pmm_management` would fix the existing PMmodule actions that gate on it and are unreachable out of the box.
- Replacing the `"r"`/`"w"` literals with `SecurityInfoManager.READ`/`WRITE` across all call sites.

---

https://claude.ai/code/session_01Wp4JBeQXptWpFucmUiqbUT


CecilFoubert:
Confirming DCO sign off for all commits

CecilFoubert:
@coderabbitai review

CecilFoubert:
@coderabbitai review

CecilFoubert:
@coderabbitai review

yingbull:
I don't have a problem with this per se, other than (and I would need to check the code to confirm) isn't there a more global approach we use elsewhere to check for privs?  And if it isn't "good enough" (it may not be!) and needs a clearer helper method, I think we need one more globally unless there is something special about this? 

Ben did some work in permissions as well, could you look this over together and confirm if you think this is the right implementation/location?  Adding Ben to review.

Ben-Heerema:
@yingbull sorry I forgot about this till now, this looks good :)

src/main/java/io/github/carlos_emr/carlos/webserv/ScheduleWs.java
```diff
@@ -146,26 +158,31 @@ public void updateAppointment(AppointmentTransfer appointmentTransfer) {
...
     }
 
     public AppointmentTransfer[] getAppointmentsUpdatedAfterDate(Date updatedAfterThisDateExclusive, int itemsToReturn, boolean useGMTTime) {
+        requirePrivilege(APPOINTMENT_OBJECT, "r");
         List<Appointment> appointments = scheduleManager.getAppointmentUpdatedAfterDate(getLoggedInInfo(), updatedAfterThisDateExclusive, itemsToReturn);
         return (AppointmentTransfer.toTransfers(appointments, useGMTTime));
     }
 
     public AppointmentArchiveTransfer[] getAppointmentArchivesUpdatedAfterDate(Date updatedAfterThisDateExclusive, int itemsToReturn, boolean useGMTTime) {
+        requirePrivilege("_appointment.UpdatedAfterDate", "x");
```
Ben-Heerema, line 179: **Blocking — wrong security object; this endpoint becomes unreachable for everyone.**

Two independent problems:

1. **Semantics inversion.** `_appointment.UpdatedAfterDate x` is an existing consent-*bypass* flag, not an access gate. Its only other use is `ScheduleManagerImpl.getAppointmentUpdatedAfterDate` (line ~273), where *absence* means "filter results by patient consent" and *presence* means "you may see unfiltered results". Requiring it here converts an opt-in widening flag into a hard deny.

2. **Not provisioned.** There is no `secObjectName` or `secObjPrivilege` row for `_appointment.UpdatedAfterDate` anywhere under `database/` (seeds, updates, or CAISI init), and I confirmed it's absent from a stock dev database. With zero rows, `OscarRoleObjectPrivilege.checkPrivilege` iterates an empty list and returns `false`, so `hasPrivilege` denies **every role, including admin**. This breaks Integrator appointment-archive sync on any stock deployment.

Suggest `requirePrivilege(APPOINTMENT_OBJECT, "r")` here, matching the sibling `getAppointmentsUpdatedAfterDate` just above, and leave the consent-bypass semantics to the manager layer.

src/main/java/io/github/carlos_emr/carlos/webserv/ProgramWs.java
```diff
@@ -51,6 +51,7 @@ public class ProgramWs extends AbstractWs {
     private ProgramManager2 programManager;
 
     public ProgramTransfer[] getAllPrograms() {
+        requirePrivilege("_pmm_management", "r");
```
Ben-Heerema, line 54: **Blocking — `_pmm_management` is not a seeded security object.**

`grep -rn "_pmm_management" database/` returns nothing (not in `oscardata.sql`, not in `updates/`, not in `caisi/initcaisi.sql`), and I confirmed it's absent from a stock dev database's `secObjectName`. With no rows, `hasPrivilege` returns `false` for every role including admin, so both `getAllPrograms()` and `getAllProgramProviders()` go from fully open to unreachable.

The object name is already used by several PMmodule 2Actions, so it's an established identifier — but those apparently rely on per-site manual provisioning. Since this PR gates previously-working SOAP endpoints on it, it should ship a `database/mysql/updates/update-YYYY-MM-DD-*.sql` migration adding the `secObjectName` row plus sensible default grants (which would incidentally fix the existing PMmodule actions on stock installs too). Note the other `_pmm*` objects that do exist (`_pmm_agencyList`, `_pmm_client.BedRoomReservation`, `_pmm_editProgram.vacancies`) don't fit these endpoints.

src/test/java/io/github/carlos_emr/carlos/webserv/AbstractWsPrivilegeTest.java
```diff
@@ -0,0 +1,437 @@
...
+import java.util.Locale;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.util.ReflectionTestUtils;
+
+@DisplayName("AbstractWs privilege enforcement")
+@Tag("unit")
+@Tag("webservice")
+class AbstractWsPrivilegeTest extends CarlosUnitTestBase {
```
Ben-Heerema, line 55: **Blocking — this test class never runs in CI.**

The Surefire `<includes>` in `pom.xml` are `**/test/**/*Test.java`, `**/tickler|messenger|provider|managers/**/*Test.java`, and `**/*UnitTest.java` / `*IntegrationTest.java` / `*RegressionTest.java`. `io/github/carlos_emr/carlos/webserv/AbstractWsPrivilegeTest` matches none of them. I verified against the CI "tests" job log for head commit `7b70111`: 8,556 tests executed, zero occurrences of `AbstractWsPrivilegeTest`. The green check does not include this file.

One-word fix: rename to `AbstractWsPrivilegeUnitTest` (matches `**/*UnitTest.java` and the `CarlosUnitTestBase` naming convention).

Relatedly, the PR description's validation section cites `LabUploadWsRegressionTest` — that file predates this PR (added in #2288) and invokes the private `importLab` via `ReflectionTestUtils.invokeMethod`, bypassing the new public-method guards entirely, so it doesn't validate this change. Please correct the description.

Two smaller test asks once it's running: a deny-path test for `getDemographics` batch reads (only the all-allowed path is covered — a regression swapping `requireReadPrivilege` for the filter helper would silently invert policy undetected), and one deny test each for the untested classes (BookingWs `bookAppointment`, one LabUploadWs upload, DocumentWs, ProgramWs, FacilityWs). Also consider one test that exercises the real `getSecurityInfoManager()` via `CarlosUnitTestBase.registerMock(SecurityInfoManager.class, ...)` instead of overriding it, so the SpringUtils wiring isn't fully untested.

src/main/java/io/github/carlos_emr/carlos/webserv/DemographicWs.java
```diff
@@ -154,7 +161,39 @@ public DemographicTransfer2[] getActiveDemographicsAfter2(@WebParam(name = "last
     }
 
 
+    private List<Demographic> filterReadableDemographics(List<Demographic> demographics) {
```
Ben-Heerema, line 164: **Important — infrastructure errors silently truncate search results.**

`SecurityInfoManagerImpl.hasPrivilege` wraps its body in `catch (Exception ex) { logger.error(...); }` and falls through to `return false` (only `PatientDirectiveException` is rethrown). Combined with this per-patient filter loop, a transient DB/Hibernate failure during any single check makes that patient **silently vanish from search results while the call still returns success** — a clinician searching a name gets a truncated list with no error signal. Fail-closed is right for a genuine deny, but "check errored" shouldn't be indistinguishable from "denied"; consider faulting the whole request on unexpected exceptions.

Two related notes on this block:
- **Cost:** each `hasPrivilege` call is ~3 DB round trips (roles + `_demographic$<id>` lookup + `_demographic` fallback) with no caching, and the manager already enforces a coarse `_demographic` read via `DemographicManagerImpl.checkPrivilege`. The per-patient pass only changes the outcome when `_demographic$<id>` override rows exist. A batch `getDemographics` call with N ids now costs ~3N extra queries; worth a coarse check up front and skipping the per-patient path when no `_demographic$*` rows exist.
- **Consistency:** searches silently filter protected patients, but `getDemographic`/`getDemographics` throw for the same patients while nonexistent IDs return null — so ID probing distinguishes "doesn't exist" from "exists but protected", which is exactly what the filtering hides. Worth picking one policy. Note also the bulk endpoints (`getActiveDemographicsAfter`/`2`) return the same PHI payload under only the global check, so a `_demographic$X` NORIGHTS override doesn't actually hide patient X from this API.

src/main/java/io/github/carlos_emr/carlos/webserv/PrescriptionWs.java
```diff
@@ -57,6 +57,7 @@ public PrescriptionTransfer getPrescription(Integer prescriptionId) {
         Prescription prescription = prescriptionManager.getPrescription(loggedInInfo, prescriptionId);
 
         if (prescription != null) {
+            requirePrivilege("_rx", "r", prescription.getDemographicId() != null ? prescription.getDemographicId().toString() : null);
```
Ben-Heerema, line 60: **Important — record is loaded before any check, and the missing-record path has no check at all.**

`PrescriptionManagerImpl.getPrescription` performs no internal `hasPrivilege` call, so line 57 is a fully unauthorized read that materializes the PHI-bearing record before this check runs. And when the prescription doesn't exist, the method returns `null` having performed no privilege check — an unprivileged caller can distinguish "prescription ID exists" (SOAP fault) from "doesn't exist" (null), an enumeration oracle over prescription IDs.

Suggest a coarse `requirePrivilege("_rx", "r")` at method entry (same shape as `getPrescriptionUpdatedAfterDate` below), keeping this demographic-scoped re-check after the load. The existing test for this method is good — it correctly pins that drug details are never fetched on deny — but please add the missing-prescription case once the entry check is in.

src/main/java/io/github/carlos_emr/carlos/webserv/FacilityWs.java
```diff
@@ -60,6 +60,7 @@ public FacilityTransfer getDefaultFacility() {
     }
 
     public FacilityTransfer[] getAllFacilities(Boolean active) {
+        requirePrivilege("_admin", "r");
```
Ben-Heerema, line 63: **Important — `_admin` is granted only to the `admin` role in stock seeds; this will break provider-role sync accounts.**

The only seeded grant is `('admin', '_admin', 'x', ...)` (`oscardata.sql:1728`). The `x` grant does satisfy `r` (verified in `hasPrivilege`'s tiering), so admins are fine — but Integrator / inter-EMR sync accounts are typically provisioned with provider-type roles and will start faulting on what is essentially a facility *directory* read. `_admin r` feels coarse for a listing endpoint; please confirm against the actual Integrator flow, and either pick a less privileged object or pair this with a migration/release note for affected service accounts.

Also: the deprecated `getDefaultFacilities()` alias above shares the deferred-policy status of `getDefaultFacility()` — worth naming it explicitly in the PR description so the follow-up covers both.

src/main/java/io/github/carlos_emr/carlos/webserv/LabUploadWs.java
```diff
@@ -70,6 +70,7 @@ public String uploadCLS(
             @WebParam(name = "contents") String contents,
             @WebParam(name = "oscar_provider_no") String oscarProviderNo
     ) {
+        requirePrivilege("_lab", "w");
```
Ben-Heerema, line 73: **Important — rollout hazard for lab feeds (the check itself is correct).**

Guard placement is right on all nine upload methods (before the try, so denials propagate as faults instead of benign `success:0` strings), and requiring `_lab w` on the *calling* account is the correct semantic since `oscar_provider_no` is the target, not the caller.

But stock seeds grant `_lab` only to the `doctor` role (`oscardata.sql:1718`). Dedicated lab-uploader service accounts with minimal custom roles will start faulting, and from the EMR's perspective lab intake just silently stops — the failure is only visible on the sender's side. This needs a release note / upgrade checklist item, ideally with a migration granting `_lab w` to lab-uploader roles.

src/main/java/io/github/carlos_emr/carlos/webserv/AbstractWs.java
```diff
@@ -65,4 +67,26 @@ protected Provider getLoggedInProvider() {
...
+
+    protected void requirePrivilege(String objectName, String privilege, String demographicNo) {
+        requirePrivilege(getLoggedInInfo(), objectName, privilege, demographicNo);
+    }
+
+    protected void requirePrivilege(LoggedInInfo loggedInInfo, String objectName, String privilege) {
+        requirePrivilege(loggedInInfo, objectName, privilege, null);
+    }
+
+    protected void requirePrivilege(LoggedInInfo loggedInInfo, String objectName, String privilege, String demographicNo) {
```
Ben-Heerema, line 87: Three notes on the helper (none blocking individually):

1. **No audit trail for denials.** WS *authentication* failures are logged via `LogAction` in `AuthenticationInWSS4JInterceptor`, but these new *authorization* denials throw with no log — and there's no CXF fault interceptor to catch them, so the only server trace is CXF's generic unwind warning. In a PIPEDA context, "provider X was denied _demographic r on patient Y" is precisely the event the audit log exists for. Suggest logging a security event (provider no, object, privilege, scope — identifiers only) before throwing.

2. **Injection.** CLAUDE.md reserves `SpringUtils.getBean()` for shims that can't be wired; every `*Ws` subclass is already an `@Component` with `@Autowired` fields, so `@Autowired protected SecurityInfoManager securityInfoManager;` works here, matches the documented pattern, and avoids a container lookup per check (which can run once per patient in the DemographicWs loops).

3. **Null session noise.** With a null `LoggedInInfo`, `SecurityInfoManagerImpl` NPEs internally, swallows it to `false`, and writes an ERROR stack trace — so every unauthenticated hit produces a misleading stack trace. An explicit null guard here that throws the SecurityException directly would keep fail-closed behavior without the noise. (Minor: `SecurityInfoManager.READ`/`WRITE` constants exist and could replace the `"r"`/`"w"` literals at call sites.)

Ben-Heerema (commented):
I had Claude give it a review using the PR review tool and it found these. It looked good at a glance a few weeks ago but these seem worth looking into. :)

src/main/java/io/github/carlos_emr/carlos/webserv/ScheduleWs.java
yingbull: Confirmed on both counts, and fixed in 548d5e0 — now `requirePrivilege(APPOINTMENT_OBJECT, "r")`, matching the sibling above.

Verification for the record:

1. **Semantics.** `ScheduleManagerImpl:274` reads `if (!hasPrivilege(..., "_appointment.UpdatedAfterDate", "x", null)) { filterProviderSpecificConsent(...) }` — absence filters, presence widens. Exactly the inversion you described.
2. **Not provisioned.** `grep -rn "_appointment.UpdatedAfterDate" database/` returns nothing. The repo also already codifies this: `SecurityObjectSeedContractUnitTest` lists it under `KNOWN_UNREACHABLE` as "absent from `secObjectName` entirely".

I added a Javadoc block on the method recording why it is *not* gated on that flag, and a test that asserts `hasPrivilege` is never called with it — so a future change back to it fails rather than silently shutting the endpoint.

---

src/main/java/io/github/carlos_emr/carlos/webserv/ProgramWs.java
yingbull: Confirmed — `_pmm_management` is granted by no migration, and it too is listed in `SecurityObjectSeedContractUnitTest`'s `KNOWN_UNREACHABLE` set. Fixed in 548d5e0.

Rather than ship a migration, I used the objects the PMmodule UI already uses for these same two screens:

- `getAllPrograms()` → `_pmm.programList r`
- `getAllProgramProviders()` → `_pmm.staffList r`

Both are granted `'x'` to `admin` **and** `doctor` in both province sets (`on/V1.0.2__on_data.sql:35906-35907, 36136-36137`; `bc/V1.0.2__bc_data.sql:41124-41125, 41354-41355`), and `x` satisfies `r` per the tiering in `hasPrivilege`. So these endpoints stay reachable on a stock install with no operator action.

That also keeps the semantics tighter than `_pmm_management` would have been — program list vs. staff list are separately grantable.

Note I did not take the migration route: repo guidance blocks Claude from writing under `database/`. The broader gap (several PMmodule 2Actions gating on `_pmm_management` and therefore unreachable out of the box) is still real and still worth a migration — but it is pre-existing debt, tracked by that allowlist entry, and wider than this PR.

---

src/test/java/io/github/carlos_emr/carlos/webserv/AbstractWsPrivilegeTest.java
yingbull: Confirmed and fixed in 548d5e0. `webserv/AbstractWsPrivilegeTest` matched none of the surefire `<includes>` (`pom.xml:1743-1753`) — renamed to `AbstractWsPrivilegeUnitTest`, which matches `**/*UnitTest.java`. It now runs: **24 tests, 0 failures**, where before it contributed zero.

Both smaller asks are in, plus the extras you suggested:

- **`getDemographics` deny path** — `shouldStopBatchLoad_whenOneDemographicIsRestricted` asserts a fault plus `verifyNoInteractions(demographicManager)`. Swapping `requireReadPrivilege` for the filter helper now fails the build.
- **Deny test per untested class** — `BookingWs.bookAppointment`, `LabUploadWs.uploadCLS`, `DocumentWs.getDocument`, `ProgramWs` (both methods, one per object), `FacilityWs.getAllFacilities`.
- **Real SpringUtils wiring** — `shouldResolveSecurityManager_throughSpringWiring` drives a service that does *not* override `getSecurityInfoManager()`, resolving through `CarlosUnitTestBase.registerMock(SecurityInfoManager.class, ...)`.

Also added while in there: a deny test for the archive endpoint, a `searchDemographicByName` deny test for the new coarse gate, and the missing-prescription case (see the `PrescriptionWs` thread).

You are right about the description — `LabUploadWsRegressionTest` predates this PR and goes through `ReflectionTestUtils.invokeMethod` on the private `importLab`, so it never touches the new public-method guards. I have rewritten the PR description's validation section accordingly.

---

src/main/java/io/github/carlos_emr/carlos/webserv/PrescriptionWs.java
yingbull: Agreed on both — fixed in 548d5e0. `getPrescription` now opens with a coarse `requirePrivilege(RX_OBJECT, "r")` before the load, keeping the demographic-scoped re-check after it (the patient scope isn't knowable until the record is in hand).

That closes the enumeration oracle: existing and missing prescription IDs are now indistinguishable to an unprivileged caller — both fault. Two tests pin it:

- `shouldStopPrescriptionLoad_whenRxReadIsMissing` — faults with `verifyNoInteractions(prescriptionManager)`, so the PHI-bearing load no longer happens ahead of the check.
- `shouldDenyIdentically_whenPrescriptionDoesNotExist` — the missing-record case you asked for, asserting the same fault.

I also hoisted the repeated `"_rx"` literals into an `RX_OBJECT` constant while touching the file.

---

src/main/java/io/github/carlos_emr/carlos/webserv/DemographicWs.java
yingbull: Partly fixed in 548d5e0; being explicit about what I did and did not do, since one of these I deliberately left.

**Cost + consistency — fixed.** Both search endpoints now take a coarse `requirePrivilege(DEMOGRAPHIC_OBJECT, "r")` before the manager call. That gives the fail-fast you suggested, and it removes the worst inconsistency: a caller holding no demographic rights at all used to get an empty list from the filter loop, indistinguishable from "no patients matched". Now they fault. `shouldStopSearch_whenDemographicReadIsMissing` pins it. `AbstractWs` also now injects `SecurityInfoManager` rather than doing a `SpringUtils.getBean` per call, which the per-patient loop was paying once per patient.

**Silent truncation on infra error — not fixed, and I don't think it can be fixed here.** You've identified the mechanism exactly: `SecurityInfoManagerImpl.hasPrivilege` catches `Exception`, logs, and falls through to `return false`, rethrowing only `PatientDirectiveException`. From inside the filter loop a thrown-and-swallowed error and a genuine deny are the same `false` — there is no signal left to branch on. Distinguishing them means changing `hasPrivilege` to let unexpected exceptions escape, which changes fail-closed behaviour for every caller in the application, not just this loop. That is a real bug but a much wider one than this PR, and not something I'd want to slip in under a SOAP RBAC change. I've written the limitation into the Javadoc on `filterReadableDemographics` so the next reader doesn't have to rediscover it. Happy to open a separate issue if you'd like it tracked.

**Skipping the per-patient path when no `_demographic$*` rows exist** — not done. It needs a cheap way to ask "do any per-patient override rows exist", which `SecurityInfoManager` doesn't currently expose, and getting the cache invalidation wrong would silently disable per-patient protection. Same reasoning: worth doing, worth doing separately.

**ID probing vs. filtering** — I kept the split deliberately, and documented why. Search filters because "your search matched a patient you cannot see" is itself the leak a `_demographic$<id>` override exists to prevent. Explicit-ID reads fault because the caller already named the patient, so a short result set would misrepresent the record rather than protect it. The residual you note is real: `getActiveDemographicsAfter`/`2` still return the full payload under only the global check, so a NORIGHTS override doesn't hide that patient from those two. That's a genuine gap and I'd rather fix it in a focused change than widen this one — flagging it explicitly so it isn't lost.

---

src/main/java/io/github/carlos_emr/carlos/webserv/AbstractWs.java
yingbull: All three addressed in 548d5e0.

**1. Audit trail.** Denials now write an `ws.accessDenied` entry via `LogAction.addLogSynchronous`, carrying object, privilege and patient scope — identifiers only, sanitized through `LogSafe`, no clinical content. The whole call is wrapped so an audit failure can never turn a denial into something else; a failed write degrades to a log warning and the `SecurityException` still throws.

**2. Injection.** Switched to `@Autowired(required = false) private SecurityInfoManager`, with `getSecurityInfoManager()` falling back to `SpringUtils` only when the field is null. I kept the accessor rather than reading the field directly for two reasons: CXF can instantiate a service outside the container, and `requirePrivilege` must never degrade to "no checker, no check"; and the existing tests override the accessor. The new `shouldResolveSecurityManager_throughSpringWiring` case covers the non-overridden path so the wiring isn't untested.

**3. Null session.** `requirePrivilege` now denies a null `LoggedInInfo` directly instead of passing it down. Same fail-closed outcome, without the misleading "Error checking privileges" stack trace on every unauthenticated hit. `shouldDenyWithoutConsultingManager_whenSessionIsMissing` asserts `verifyNoInteractions(securityInfoManager)`.

On the `SecurityInfoManager.READ`/`WRITE` constants: I left the `"r"`/`"w"` literals as-is. Swapping them touches every call site across all nine services for no behavioural change, which would bury the substantive fixes in this diff. Happy to do it as a follow-up sweep if you'd prefer it landed.

---

src/main/java/io/github/carlos_emr/carlos/webserv/FacilityWs.java
yingbull: Confirmed your reading of the seeds — in the current Flyway set the only `_admin` grant is `('admin','_admin','x',...)` in both provinces, and `x` satisfies `r`. So admin passes; provider-role sync accounts do not.

I took your option (b) rather than (a). I could not confirm the Integrator flow from this repository — Integrator is an external caller and nothing here pins which role its service accounts run under — and picking a different object would have traded one unverified policy for another. Removing the guard outright I liked less: it's a live hardening on an endpoint that's currently wide open.

So `_admin r` stays, and the rollout cost is now documented rather than discovered in production:

- Method-level Javadoc on `getAllFacilities` stating the requirement and the upgrade impact.
- A new `docs/soap-rbac-hardening.md` with the full per-endpoint privilege map and an upgrade checklist — item 2 is exactly this: inventory Integrator/inter-EMR accounts, grant `_admin r` or move them to an admin-role account.
- The checklist also notes that after deploying, `ws.accessDenied` audit entries (new in this PR) name the object and privilege refused, so any integration that needs a grant shows up by name rather than as a silent failure.

If you know what role the Integrator accounts actually use, I'll happily narrow the object instead — that's the better fix and I'd rather make it than document around it.

Good catch on `getDefaultFacilities()` — the deprecated alias is now named alongside `getDefaultFacility()` in the deferred list, both in the PR description and in the doc's Deferred section.

---

src/main/java/io/github/carlos_emr/carlos/webserv/LabUploadWs.java
yingbull: Thanks — and agreed the hazard is the real risk here, precisely because the EMR side looks fine while intake stops. Confirmed the seed situation: `_lab` is granted only to `doctor`, in both province sets.

Release note added in 548d5e0:

- Class-level Javadoc on `LabUploadWs` recording that the check is on the *calling* account (not `oscar_provider_no`), why it sits before the try block, and the upgrade impact.
- `docs/soap-rbac-hardening.md` item 1 in the upgrade checklist: inventory the accounts used by lab feeds and grant `_lab w` to their roles before deploying.

I did not add the migration granting `_lab w` to lab-uploader roles — there's no stock "lab uploader" role to grant it to, so the target is site-specific, and repo guidance blocks Claude from writing under `database/`. The checklist frames it as an operator step for that reason.

One thing that makes this less silent than before: the new `ws.accessDenied` audit entries name the object and privilege refused, so after deploying, every service account that needs a grant appears in the audit log by provider number rather than failing invisibly. That's called out at the end of the checklist.

Your point about placement-before-the-try is now pinned by a test — `shouldRejectUpload_whenLabWriteIsMissing` asserts the `SecurityException` escapes rather than being flattened into the `success:0` JSON body, which a sender would read as a routine per-file rejection and never retry.

---

src/main/java/io/github/carlos_emr/carlos/webserv/DemographicWs.java
yingbull: Following up on my own note here — the residual I flagged (`getActiveDemographicsAfter`/`2` returning the full payload under only the global check) is now **fixed** in 860eb81, not just documented.

Both `getActiveDemographicsAfter` and `getActiveDemographicsAfter2` route their results through the same `filterReadableDemographics` helper the search endpoints use, so a `_demographic$X` NORIGHTS override now hides patient X from the bulk exports too. Two tests pin it (`shouldFilterUnreadablePatients_fromActiveDemographicExport` and the `...Version2` variant).

Credit where due: `chatgpt-codex-connector` and `copilot-pull-request-reviewer` both independently flagged this same gap, which is what prompted me to fix rather than defer it. Turning to fix it surfaced the same bug in three sibling services — see the new threads on `DocumentWs`, `MeasurementWs` and `PrescriptionWs`.

Still not fixed, and still needing a decision rather than a patch: the `hasPrivilege` infrastructure-error swallowing. That remains exactly as I described above — there's no signal left in the loop to branch on, and changing it alters fail-closed behaviour application-wide. Say the word and I'll open the tracking issue.

---

src/main/java/io/github/carlos_emr/carlos/webserv/DocumentWs.java
```diff
@@ -76,19 +78,22 @@ public DocumentTransfer getDocument(Integer documentId) {
...
     public DocumentTransfer[] getDocumentsByProgramProviderDemographicDate(Integer programId, String providerNo, Integer demographicId, Calendar updatedAfterThisDateExclusive, int itemsToReturn) {
+        requirePrivilege(EDOC_OBJECT, "r");
         LoggedInInfo loggedInInfo = getLoggedInInfo();
         List<Document> documents = documentManager.getDocumentsByProgramProviderDemographicDate(loggedInInfo, programId, providerNo, demographicId, updatedAfterThisDateExclusive, itemsToReturn);
         logger.debug("programId=" + programId + ", providerNo=" + providerNo + ", demographicId=" + demographicId + ", updatedAfterThisDateExclusive=" + DateFormatUtils.ISO_DATETIME_FORMAT.format(updatedAfterThisDateExclusive) + ", itemsToReturn=" + itemsToReturn + ", results=" + documents.size());
         return (DocumentTransfer.getTransfers(loggedInInfo, documents));
     }
 
     public DocumentTransfer[] getDocumentsByDemographicIdAfter(@WebParam(name = "lastUpdate") Calendar lastUpdate, @WebParam(name = "demographicId") Integer demographicId) {
+        requirePrivilege(EDOC_OBJECT, "r");
```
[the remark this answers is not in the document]
yingbull: Verified and fixed in 860eb81 — a real bypass, thank you.

`getDocumentsByDemographicIdAfter` and `getDocumentsByProgramProviderDemographicDate` both have the patient in hand and now pass it to `requirePrivilege`.

The sibling you didn't have to spell out was harder: `getDocument(documentId)` has no patient available before the load. It now re-checks after resolving the `ctl_document` row. That row is a generic join table, so `moduleId` only means a demographic number when `module` is `demographic` — anything else stays on the unscoped decision rather than being misread as a patient id.

`getDocumentsUpdateAfterDate` filters per record instead of throwing, matching the demographic searches: throwing on the first restricted hit would confirm the restricted patient has documents in the window, which is what the override exists to hide. That costs a second `ctl_document` lookup per document on top of the one `DocumentTransfer.getTransfers` already performs — deliberate, since the transfer embeds the file bytes and the decision must be made before it runs.

Covered by `shouldPassPatientScope_whenReadingDocumentsForDemographic`.

---

src/main/java/io/github/carlos_emr/carlos/webserv/MeasurementWs.java
```diff
@@ -71,18 +76,21 @@ public MeasurementMapTransfer[] getMeasurementMaps() {
...
     public Integer addMeasurement(MeasurementTransfer measurementTransfer) {
+        requirePrivilege(MEASUREMENT_OBJECT, "w", measurementTransfer.getDemographicId() != null ? measurementTransfer.getDemographicId().toString() : null);
         Measurement measurement = new Measurement();
         measurementTransfer.copyTo(measurement);
         measurementManager.addMeasurement(getLoggedInInfo(), measurement);
         return (measurement.getId());
     }
 
     public MeasurementTransfer[] getMeasurementsByProgramProviderDemographicDate(Integer programId, String providerNo, Integer demographicId, Calendar updatedAfterThisDateExclusive, int itemsToReturn) {
+        requirePrivilege(MEASUREMENT_OBJECT, "r");
```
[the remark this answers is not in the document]
yingbull: Verified and fixed in 860eb81, including the "other demographic-specific endpoint" you pointed at.

- `getMeasurementsByDemographicIdAfter` and `getMeasurementsByProgramProviderDemographicDate` now pass the requested `demographicId` to `requirePrivilege`.
- `getMeasurement(measurementId)` keeps the unscoped gate and adds a scoped re-check on the loaded record's `demographicId`. The patient isn't knowable before the load and the manager performs no check of its own, so that ordering is forced; what matters is that the transfer is never built for a patient the caller can't read.
- `getMeasurementsCreatedAfterDate` filters per record rather than throwing, for the same reason as the demographic searches.

Three tests cover it: `shouldEnforceLoadedRecordScope_whenReadingMeasurement`, `shouldFilterUnreadablePatients_fromMeasurementSync`, and `shouldPassPatientScope_whenReadingMeasurementsForDemographic`.

`addMeasurement` was already demographic-scoped from an earlier round of this PR.

---

src/main/java/io/github/carlos_emr/carlos/webserv/DemographicWs.java
```diff
@@ -107,11 +111,13 @@ public DemographicTransfer[] getDemographics(Integer[] demographicIds) {
             ids.add(i);
         }
 
+        requireReadPrivilege(ids);
         List<Demographic> demographics = demographicManager.getDemographics(getLoggedInInfo(), ids);
         return (DemographicTransfer.toTransfers(demographics));
     }
 
     public DemographicTransfer[] getActiveDemographicsAfter(@WebParam(name = "lastUpdate") Calendar lastUpdate, @WebParam(name = "fields") String fields) {
+        requirePrivilege(DEMOGRAPHIC_OBJECT, "r");
```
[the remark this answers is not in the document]
yingbull: Verified and fixed in 860eb81. Both `getActiveDemographicsAfter` and `getActiveDemographicsAfter2` now route their results through the same `filterReadableDemographics` helper the search endpoints use, so a `_demographic$X` NORIGHTS override hides patient X from the bulk exports as well.

This one was already known — I'd flagged it in the human review thread and listed it under PR follow-up — so your report is what moved it from "documented gap" to "fixed". Chasing it surfaced the same bug in `DocumentWs`, `MeasurementWs` and `PrescriptionWs`, which are fixed in the same commit.

Pinned by `shouldFilterUnreadablePatients_fromActiveDemographicExport` and the `...Version2` variant.

---

src/main/java/io/github/carlos_emr/carlos/webserv/PrescriptionWs.java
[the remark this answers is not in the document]
yingbull: Both PrescriptionWs findings are fixed, in two separate commits.

**Entry check** (this thread): fixed earlier in 548d5e0 — `getPrescription` now opens with an unscoped `requirePrivilege(RX_OBJECT, "r")` before the manager call, keeping the demographic-scoped re-check after the load. That closes the ID-probing oracle you describe: existing and missing prescription IDs now produce the same fault. `shouldStopPrescriptionLoad_whenRxReadIsMissing` and `shouldDenyIdentically_whenPrescriptionDoesNotExist` pin both halves.

**Bulk endpoint** (your second comment): fixed in 860eb81 — `getPrescriptionUpdatedAfterDate` filters each result by its demographic privilege before `PrescriptionTransfer.getTransfers`, which is where drug detail gets loaded. Filtering rather than throwing, so a restricted patient's presence in the window isn't revealed by a fault. `shouldFilterUnreadablePatients_fromPrescriptionSync` asserts the readable prescription still reaches `getDrugsByScriptNo` and the restricted one never does — an over-broad filter fails the test as loudly as a missing one.

I kept the enforcement in the Ws layer rather than the manager, as you offered: the manager is shared with callers that have their own authorization paths, and widening it is a larger change than this PR should carry.

---

src/main/java/io/github/carlos_emr/carlos/webserv/DemographicWs.java
```diff
@@ -107,11 +111,13 @@ public DemographicTransfer[] getDemographics(Integer[] demographicIds) {
...
         Date afterDateExclusive = lastUpdate != null ? lastUpdate.getTime() : null;
         List<Demographic> demographics = demographicManager.getActiveDemographicAfter(getLoggedInInfo(), afterDateExclusive);
```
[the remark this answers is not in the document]
yingbull: Fixed in 860eb81, including line 142 — both `getActiveDemographicsAfter` and `getActiveDemographicsAfter2` now filter with `filterReadableDemographics` before converting, exactly as you suggested and matching the search methods.

Pinned by `shouldFilterUnreadablePatients_fromActiveDemographicExport` and the `...Version2` variant.

---

src/main/java/io/github/carlos_emr/carlos/webserv/DocumentWs.java
```diff
@@ -64,6 +65,7 @@ public class DocumentWs extends AbstractWs {
     private ProgramManager programManager;
 
     public DocumentTransfer getDocument(Integer documentId) {
+        requirePrivilege(EDOC_OBJECT, "r");
```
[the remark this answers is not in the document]
yingbull: Fixed in 860eb81 across all four locations you listed.

Your suggestion to "resolve the ctl document's demographic module ID and enforce the scoped privilege before building the transfer" is exactly what `getDocument` now does — and you were right that it matters specifically because `DocumentTransfer.toTransfer` embeds the file bytes.

One refinement worth noting: `ctl_document` is a generic join table, so `moduleId` only means a demographic number when `module` is `demographic`. The helper returns null for any other module rather than treating the value as a patient id, and those rows stay on the unscoped decision.

Lines 81 and 96 (`getDocumentsUpdateAfterDate`, `getDocumentsByDemographicIdAfter`) and `getDocumentsByProgramProviderDemographicDate` are all covered too — the ones with a patient parameter pass it directly; the bulk one filters per record.

---

src/main/java/io/github/carlos_emr/carlos/webserv/MeasurementWs.java
```diff
@@ -49,20 +49,25 @@
 @Component
 @GZIP(threshold = AbstractWs.GZIP_THRESHOLD)
 public class MeasurementWs extends AbstractWs {
+    private static final String MEASUREMENT_OBJECT = "_measurement";
+
     @Autowired
     private MeasurementManager measurementManager;
 
     public MeasurementTransfer getMeasurement(Integer measurementId) {
+        requirePrivilege(MEASUREMENT_OBJECT, "r");
```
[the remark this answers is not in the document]
yingbull: Fixed in 860eb81 across all four locations.

`getMeasurement` follows your suggested shape precisely: the global gate is retained before the lookup, then the loaded measurement's `demographicId` is enforced before the transfer is built. Lines 87 and 93 take the requested `demographicId` directly, and `getMeasurementsCreatedAfterDate` filters per record.

`getMeasurementMaps` is left unscoped deliberately — it returns measurement-type reference data, not patient records.

Three tests cover it: `shouldEnforceLoadedRecordScope_whenReadingMeasurement`, `shouldFilterUnreadablePatients_fromMeasurementSync`, `shouldPassPatientScope_whenReadingMeasurementsForDemographic`.

---

Issue #2953, opened 2026-06-19
getAppointment 500 response leaks patient PHI in body (claude assist)


Ben-Heerema:
`POST /ws/rs/schedule/getAppointment` returns **HTTP 500** for a valid appointment, but the response **body still contains the full appointment + nested demographic JSON** (patient name, phone, patient status). Confirmed live.

**Root cause / why it leaks:** the 500 is a mid-stream Jackson serialization failure, so partial JSON is already written. `ResponseSanitizationFilter` only blanks bodies that contain stack-trace markers — a clean JSON error body passes through unsanitized.

**Fix direction:** ensure error responses (esp. 5xx) on `/ws` do not emit partial entity bodies, and/or have the sanitizer suppress bodies on error status regardless of stack-trace content.

**Suggested type:** Security · **Priority:** 2-High (PHI exposure)
Related: schedule endpoint 500s tracked separately.

---
Found during the API testing pass in #2916 (verified against `develop`, 2026-06-19). Filed with Claude assistance.

PR #2985, opened 2026-06-22
fix: suppress web-service 5xx response bodies to stop PHI leak (#2953)

Ben-Heerema:
## Summary

Relates to #2953 — `POST /ws/rs/schedule/getAppointment` returned **HTTP 500** while the response body still contained the partially-serialized appointment and nested demographic JSON (patient name, phone, patient status), leaking PHI.

**Root cause:** a mid-stream Jackson serialization failure leaves a clean, well-formed JSON prefix already written before the failure. `ResponseSanitizationFilter` only replaced bodies that matched its **stack-trace** heuristic, so a clean JSON error body passed through unsanitized (CWE-209).

## Fix

Replace the body of **any 5xx response on a web-service route** (`/ws/*`, the CXF servlet) regardless of stack-trace content, on both the captured writer and output-stream error paths.

- `sanitizationReason(status, body, webServiceRequest)` — central decision: returns a log-safe reason code (`stack-trace-marker` for any route with stack-trace markers; `web-service-5xx-partial-body` for `/ws/*` with status ≥ 500) or `null` to pass through. Computed once per error response.
- `shouldSanitizeErrorBody(...)` — thin predicate over the above (retained for callers/tests).
- `isWebServiceRequest(request)` — matches the CXF servlet path (`/ws/*` per `web.xml`) using the servlet path, with a **context-relative** URI fallback (exact/prefix match, not a substring, to avoid false positives like `/proxy/ws/foo`).

Scoped deliberately, per the response-filter safety guidance in `CLAUDE.md` (targeted by route + status):

- **4xx on `/ws`** keep the stack-trace-only check, so legitimate REST JSON error envelopes (validation messages, `401`/`403`/`404`) are preserved.
- **Non-`/ws` routes** are unaffected — a normal 500 HTML page without a stack trace still passes through.
- The output-stream *passthrough* path already blanked any uncommitted `≥400`; this aligns the captured paths with that behaviour.

> ⚠️ **Behaviour note for reviewers:** on `/ws`, *any* 5xx body is now replaced with the generic HTML error page — a deliberate security-over-contract tradeoff. If a `/ws` endpoint intentionally returns a machine-readable JSON body on 5xx, call it out here. 4xx and non-`/ws` are untouched. (See inline reviewer note comment.)

## Tests

`ResponseSanitizationFilterUnitTest` — **66 tests, all passing**:

- `/ws` 500 partial-JSON PHI leak via **writer** and **output stream** → body replaced with the generic Reference-ID page; PHI absent.
- `/ws` 500 carrying a stack trace → sanitized end-to-end via the stack-trace trigger.
- `/ws` 400 JSON error envelope and `/ws` 200 JSON body → passed through unchanged.
- non-`/ws` 500 without a stack trace → passed through unchanged.
- `sanitizationReason()` / `shouldSanitizeErrorBody()` decision matrix and `isWebServiceRequest()` (incl. a `/proxy/ws/foo` false-positive guard).

Also validated with a live embedded-Jetty A/B (real container, real HTTP): the writer and captured-output-stream `/ws` 5xx paths leak on `develop` and are blanked with this PR.

## Known limitation / follow-up

This fix covers cases where the response is **not yet committed**. If a `/ws` 5xx partial body exceeds the container output buffer, the bytes are flushed before the filter can intervene and PHI still leaks — tracked separately in **#2994** (output-buffering / CXF interceptor). Out of scope here.

### DIFF
--- a/src/main/java/io/github/carlos_emr/carlos/utility/ResponseSanitizationFilter.java
+++ b/src/main/java/io/github/carlos_emr/carlos/utility/ResponseSanitizationFilter.java
@@ -317,14 +317,15 @@ public class ResponseSanitizationFilter implements Filter {
             int status = wrapper.getStatus();
             byte[] capturedBytes = wrapper.getCapturedOutput();
             String capturedBody = new String(capturedBytes, StandardCharsets.UTF_8);
-            if (shouldSanitizeErrorBody(status, capturedBody, webServiceRequest)) {
+            String reason = sanitizationReason(status, capturedBody, webServiceRequest);
+            if (reason != null) {
                 String correlationId = generateCorrelationId();
                 LOGGER.error("Sanitizing output-stream error response body "
                                 + "[status={} uri={} correlationId={} reason={}]",
                         status,
                         LogSafe.sanitizeUri(((HttpServletRequest) request).getRequestURI()),
                         correlationId,
-                        sanitizationReason(capturedBody));
+                        reason);
                 sendSanitizedError(httpResponse, status, correlationId);
             } else {
                 writeBytesToResponse(httpResponse, capturedBytes);
@@ -348,7 +349,8 @@ public class ResponseSanitizationFilter implements Filter {
         int status = wrapper.getStatus();
         String capturedBody = wrapper.getCapturedContent();
 
-        if (shouldSanitizeErrorBody(status, capturedBody, webServiceRequest)) {
+        String reason = sanitizationReason(status, capturedBody, webServiceRequest);
+        if (reason != null) {
             // Tainted (stack trace) or web-service 5xx partial body: log correlation details
             // only and send a sanitized replacement.
             String correlationId = generateCorrelationId();
@@ -357,7 +359,7 @@ public class ResponseSanitizationFilter implements Filter {
                     status,
                     LogSafe.sanitizeUri(((HttpServletRequest) request).getRequestURI()),
                     correlationId,
-                    sanitizationReason(capturedBody));
+                    reason);
             sendSanitizedError(httpResponse, status, correlationId);
         } else {
             // Safe response — write captured content through to the real response.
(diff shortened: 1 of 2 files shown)


Ben-Heerema:
@coderabbitai review
@cubic-dev-ai review

Ben-Heerema:
@coderabbitai review
@cubic-dev-ai review

Ben-Heerema:
### Reviewer note: behaviour change for `/ws` 5xx responses **with** a body

Flagging a deliberate scope decision for awareness during review.

With this change, on web-service routes (`/ws/*`) **any** 5xx response body is replaced with the generic HTML "Reference ID" error page, regardless of content — not only bodies containing stack-trace markers. This is the intended security-over-contract tradeoff that closes the PHI leak in #2953 (a mid-stream Jackson failure emits clean JSON with no stack trace, which the old heuristic could not catch).

Implications:
- If a `/ws` endpoint **intentionally** returns a 500 with a structured JSON error envelope, a JSON-expecting client will now receive `text/html` instead. A genuine 500 is already a failed request, and these endpoints carry no client-meaningful entity body on 5xx, so this is considered acceptable.
- **4xx is unaffected** — legitimate REST client-error envelopes (validation messages, `401`/`403`/`404`) still pass through untouched.
- Non-`/ws` routes are unaffected.

If any `/ws` endpoint relies on a machine-readable 5xx JSON body as part of its contract, that should be called out here before merge.

Ben-Heerema:
@coderabbitai review
@cubic-dev-ai review

Ben-Heerema:
Follow-up filed: #2994 tracks the remaining **committed-response** gap — a `/ws` 5xx whose partial body exceeds the container output buffer is flushed before this filter can intervene, so PHI still leaks in that case (confirmed via an embedded-Jetty A/B). This PR closes the uncommitted writer / captured-output-stream paths; the buffering fix is scoped to #2994.

Ben-Heerema:
@coderabbitai review

Ben-Heerema:
@cubic-dev-ai review
@coderabbitai review

yingbull (approved):

PR #3077, opened 2026-06-28
test: assert forced-reset staging keeps credential material out of session (fixes #2301)

### LINKED ISSUE #2301: test: add LoginCheckLoginBean unit tests for null-session, timing-equalization, and forced-reset paths (claude assist)
## Summary

\`LoginCheckLoginBean\` (the core authentication bean) has no dedicated unit tests. The browser-level Playwright script (\`scripts/login-playwright-checks.js\`, tracked in #2265) covers the visible login flow, and PR #767 adds \`SecurityManagerUnitTest\` for password encoding and record management — but neither covers the Java logic inside \`LoginCheckLoginBean\` itself.

### 1. Timing equalization for missing users

\`MISSING_USER_DUMMY_PASSWORD_HASH\` is used to ensure BCrypt runs even when a username isn't found, preventing timing-based username enumeration. A test should confirm: when the username doesn't exist, \`securityManager.matches(dummyHash, submittedPassword)\` is called and the method returns \`false\` (not an exception or a short-circuit that would leak timing information).

### 2. Null / missing session fields

\`LoginCheckLoginBean\` reads \`userrole\` and \`user\` from the session. A test should confirm that a null or absent session attribute results in a failed authentication, not a \`NullPointerException\` surfacing to the caller.

### 3. Forced-reset path does not store credential material in session

Per the invariants documented in \`CLAUDE.md\`: forced-reset credential material stays out of the HTTP session and is referenced by an opaque, short-lived cache token. A unit test should verify that after a forced-reset path is entered, no raw password or hash appears in any session attribute.

## Suggested structure

\`\`\`java
@Tag("unit")
@Tag("login")
@Tag("security")
class LoginCheckLoginBeanUnitTest extends CarlosUnitTestBase {

    @Nested
    @DisplayName("when username does not exist")
    class WhenUsernameNotFound { ... }

    @Nested
    @DisplayName("when session attributes are missing")
    class WhenSessionMissing { ... }

    @Nested
    @DisplayName("forced-reset credential isolation")
    class ForcedResetCredentialIsolation { ... }
}
\`\`\`

## Related

- #2265 — browser-level login/logout Playwright tests (different layer)
- PR #767 — \`SecurityManagerUnitTest\` (password encoding/record management, not the bean itself)

AnuragJCChaturvedi:
## What

Adds a unit test to `Login2ActionForcedPasswordResetUnitTest` that pins the
forced-reset **credential-isolation invariant**: during forced-password-reset
staging, no credential material is written to the HTTP session.

The test invokes `setUserInfoToSession`, then:
- enumerates **every** session attribute and asserts none contains the raw
  password, the password hash, or the PIN;
- excludes only the opaque cache token from the PIN substring scan, since the
  256-bit Base64URL token can coincidentally contain a 4-digit sequence (the
  token is separately asserted to be a non-credential opaque reference, and the
  PIN cannot enter it — tokens come from `SecureRandom`, independent of the PIN);
- confirms the credentials — including the PIN — round-trip through the
  server-side `LoginCredentialCache` via the token.

## Why

This maps directly to the CLAUDE.md login invariant: *"forced-reset credential
material stays out of the HTTP session and is referenced by an opaque,
short-lived cache token."* The existing suite covered the credential-token
lifecycle (mint / invalidate / consume) but never asserted the isolation
invariant itself — so a regression that placed the password, hash, or PIN back
into the session would have gone undetected.

No production code changes; test-only.

## Testing

mvn test -Dtest=Login2ActionForcedPasswordResetUnitTest

## Related

Closes #2301

Enhanced with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Tests:
- Add a unit test that asserts no password, password hash, or PIN is stored in the HTTP session during forced-reset staging and that credentials are only retrievable via the opaque token-backed cache.

---
## Summary by cubic
Add a unit test ensuring forced-password-reset staging keeps credentials out of the HTTP session and uses only an opaque token with `LoginCredentialCache`, fixing #2301. It asserts a session is created, scans every session attribute (decoding char[]/byte[]) for leaks, verifies credentials round-trip via the token, and cleans up the cache in a finally block.

<sup>Written for commit e27338cf1c0b4091b779cd4168ac852065c3e182. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3077?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/test/java/io/github/carlos_emr/carlos/login/Login2ActionForcedPasswordResetUnitTest.java
+++ b/src/test/java/io/github/carlos_emr/carlos/login/Login2ActionForcedPasswordResetUnitTest.java
@@ -1916,6 +1916,56 @@ class Login2ActionForcedPasswordResetUnitTest extends CarlosUnitTestBase {
     }
 
     @Test
+    @DisplayName("should keep credential material out of session when staging forced reset")
+    void shouldKeepCredentialMaterialOutOfSession_whenStagingForcedReset() {
+        String password = VALID_PASSWORD;
+        String pin = "2026";
+        when(securityManager.encodePassword(password)).thenReturn(ENCODED_OLD_PASSWORD);
+        Login2Action action = newAction(null, null, null);
+
+        ReflectionTestUtils.invokeMethod(action, "setUserInfoToSession", request,
+                USERNAME, password, pin, "/provider/providercontrol.jsp");
+
+        MockHttpSession session = (MockHttpSession) request.getSession(false);
+        String token = (String) session.getAttribute(Login2Action.LOGIN_CREDENTIALS_TOKEN_ATTR);
+
+        // The session carries only the opaque token, which is not itself credential material.
+        assertThat(token).isNotBlank()
+                .doesNotContain(password)
+                .doesNotContain(ENCODED_OLD_PASSWORD);
+
+        // Strong form: enumerate EVERY session attribute and assert none leaks the raw
+        // password, the password hash, or the PIN, so a regression that stashed credential
+        // material under any other key would fail here (not just the known token attribute).
+        Collections.list(session.getAttributeNames()).forEach(name -> {
+            String value = String.valueOf(session.getAttribute(name));
+            assertThat(value)
+                    .as("session attribute '%s' must not contain the password or hash", name)
+                    .doesNotContain(password)
+                    .doesNotContain(ENCODED_OLD_PASSWORD);
+            // The opaque token is a 256-bit Base64URL random reference (verified above), not
+            // credential material; a 4-digit PIN can appear inside it by chance, so the PIN
+            // substring scan is applied to every OTHER attribute only. The PIN cannot leak
+            // into the token itself — tokens come from SecureRandom, independent of the PIN.
+            if (!Login2Action.LOGIN_CREDENTIALS_TOKEN_ATTR.equals(name)) {
+                assertThat(value)
+                        .as("session attribute '%s' must not contain the PIN", name)
+                        .doesNotContain(pin);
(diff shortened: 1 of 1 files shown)


AnuragJCChaturvedi:
Hi @yingbull. I've reworked the test into the strong form. It now enumerates every session attribute and asserts none leaks the password, hash, or PIN, rather than just checking the token key.

I also folded in the automated-review suggestions that looked legitimate:
- decode attribute values by real type (char[]/byte[]) instead of String.valueOf, so array-typed secrets aren't missed;
- guard against a null session before reading attributes;
 - invalidate the cache entry in a finally so a failed assertion can't leak state into other tests.
 
One deliberate choice: the opaque token is excluded from the PIN substring scan only, a random Base64URL token can contain a 4-digit run by chance, and there's no path for the PIN to reach the token (it's SecureRandom-generated). Happy to change that if you'd prefer it scanned too.

DCO/CI is green now. Would appreciate a review when you get a chance, happy to answer any questions.

yingbull (approved):

yingbull:
Thanks for the contribution! A bit slammed recently with a) many awesome contributions and b) a release, but the gratitude remains even if this was slow.  Thank you!

PR #3079, opened 2026-06-28
fix: preserve chart label fallback on path security errors

### LINKED ISSUE #3065: bug: PrintDemoChartLabel2Action label load does not catch SecurityException from resolveTrustedPath (skips fallback)
## Summary

`PrintDemoChartLabel2Action`'s label-file load only catches `FileNotFoundException`, but `PathValidationUtils.resolveTrustedPath(...)` can also throw `SecurityException`. If path resolution fails with a `SecurityException`, it escapes the inner try and the classpath fallback (`/oscar/oscarDemographic/<labelFile>`) is skipped, so the action errors instead of degrading gracefully. Sibling actions `PrintDemoLabel2Action` and `PrintDemoAddressLabel2Action` already catch both.

## Type
Bug

## Priority
3-Medium — inconsistent error handling; a path-resolution failure becomes an unhandled error instead of the intended graceful fallback.

## Kind
Task

## Related
Surfaced during review of #2548 (constructor-injection migration); pre-existing, outside that PR's diff. Parent: #2423.

## Affected File / Location
`src/main/java/io/github/carlos_emr/carlos/demographic/PrintDemoChartLabel2Action.java`

```java
// ~line 226-235  (only FileNotFoundException is caught)
ins = new FileInputStream(PathValidationUtils.resolveTrustedPath(
        new File(System.getProperty("user.home") + File.separator + labelFile)));
} catch (FileNotFoundException ex1) {       // <-- SecurityException escapes
    ...
    ins = getClass().getResourceAsStream("/oscar/oscarDemographic/" + labelFile);
```

Compare `PrintDemoLabel2Action` (~line 219), which does:
```java
} catch (FileNotFoundException | SecurityException ex1) {
```

## Fix
Catch `FileNotFoundException | SecurityException` in the label-load fallback in `PrintDemoChartLabel2Action`, matching the sibling print-label actions, so a path-resolution failure falls back to the classpath resource and the direct-response action still returns `NONE`.

## Acceptance Criteria
- A `SecurityException` from `resolveTrustedPath` falls back to the classpath label resource (no escape).
- Behaviour matches `PrintDemoLabel2Action` / `PrintDemoAddressLabel2Action`.

DragonFSKY:
## Summary
- Treat `SecurityException` like `FileNotFoundException` when loading the user-home chart label.
- Preserve the existing classpath fallback, matching the sibling print-label actions.

Fixes #3065

## Testing
- `mvn -Dtest=ViewPrintDemographicLabelAuditLoggingUnitTest test` (4 tests passed)
- `git diff --check`

## Summary by Sourcery

Restore reliable chart-label fallback behavior for inaccessible user-home templates.

Bug Fixes:
- Preserve the classpath chart-label fallback when the user-home label file cannot be accessed due to a security restriction.

Tests:
- Add unit coverage confirming chart-label generation falls back to the classpath template after a path security error.

### DIFF
--- a/src/main/java/io/github/carlos_emr/carlos/demographic/PrintDemoChartLabel2Action.java
+++ b/src/main/java/io/github/carlos_emr/carlos/demographic/PrintDemoChartLabel2Action.java
@@ -226,6 +226,8 @@ public class PrintDemoChartLabel2Action extends ActionSupport {
                 ins = new FileInputStream(PathValidationUtils.resolveTrustedPath(new File(System.getProperty("user.home") + File.separator + labelFile)));
             } catch (FileNotFoundException ex1) {
                 logger.warn(labelFile + " not found in user's home directory. Using default instead (classpath)", ex1);
+            } catch (SecurityException ex1) {
+                logger.warn(labelFile + " is not accessible from user's home directory. Using default instead (classpath)", ex1);
             }
 
             if (ins == null) {
--- a/src/test/java/io/github/carlos_emr/carlos/demographic/PrintDemoChartLabel2ActionUnitTest.java
+++ b/src/test/java/io/github/carlos_emr/carlos/demographic/PrintDemoChartLabel2ActionUnitTest.java
@@ -1,0 +1,126 @@
+/**
+ * Copyright (c) 2026 CARLOS Contributors. All Rights Reserved.
+ *
+ * This software is published under the GPL GNU General Public License.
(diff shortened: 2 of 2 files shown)


DragonFSKY:
Thanks for the review. I split the FileNotFoundException and SecurityException handling so the fallback behavior stays the same, while the warning now distinguishes an inaccessible user-home label path from a missing file. The focused label test passes locally.

Issue #2906, opened 2026-06-15
Security: audit request-sourced identity parameters that should come from the session


yingbull:
Follow-up from the `document-upload` PR, which removed the caller-supplied `curUser` parameter from `ViewDocumentReport` URLs and now derives the current user from the session. Other endpoints still read an *identity-of-the-acting-user* value from the request where it should instead be the authenticated session provider. When the value drives a "creator"/"acting user" field, an audit/ownership record, or a billing submitter, trusting the request allows spoofing.

### Closest analogs to the PR (highest priority — `curUser`/creator from request)
- `billings/ca/on/web/ViewInrBillingGeneration2Action.java:113` — `String creator = request.getParameter("curUser")`
- `billings/ca/on/web/ViewOnInrBillingGeneration2Action.java:110` — `String creator = request.getParameter("curUser")`
- `billings/ca/on/service/OhipReportGenerationService.java:234` — `String curUser = request.getParameter("curUser")`
- `billings/ca/bc/pageUtil/BillingBCSetup2Action.java:158` — `bean.setCreator(request.getParameter("user_no"))`
- `tickler/pageUtil/DbTicklerAdd2Action.java:124` — `docCreator` from `request.getParameter("user_no")`

### Second tier — `provider_no`/`providerNo` written as ownership/creator (review case-by-case)
Some of these are legitimately admin-selected (e.g. assigning an appointment to another provider) and should keep the parameter but add an authorization check. Examples: `AppointmentAddRecord2Action`, `DemographicUpdate2Action`, `SecurityAddSecurityHelper`, `BillingCorrectionService`, `EctIncomingEncounter2Action`. Distinguish *acting-user identity* (must be session) from *selected subject provider* (param OK, but must be `hasPrivilege`/`isAllowedAccessToPatientRecord`-checked).

### Acceptance criteria
- [ ] Re-run the search (the list above is a non-exhaustive read-only scan) and triage each hit.
- [ ] For each site, classify as *acting-user identity* (→ derive from `LoggedInInfo`/session) vs *selected subject* (→ keep param but enforce authorization).
- [ ] Replace acting-user reads with session-derived values; add a regression assertion per fix.
- [ ] Document any intentional admin-on-behalf-of cases.

### Classification
Type: **Security** · Priority: **3-Medium** · Kind: **Task**

_Generated as a follow-up to the document-upload security review._

yingbull:
Consolidating #2855 here.

[...]

PR #3080, opened 2026-06-28
fix: derive INR billing creator from session

DragonFSKY:
## Summary
- Derive the creator from the authenticated session provider in both INR billing generation paths.
- Update the existing tests to verify a caller-supplied `curUser` is ignored.

Partially addresses #2906.

## Testing
- `mvn -Dtest=ViewOnInrBillingGeneration2ActionUnitTest,ViewInrBillingGeneration2ActionUnitTest test` (16 tests passed)
- `git diff --check`


DragonFSKY:
Thanks for the review. I updated this to fail fast when the session provider number is missing or blank, and added regression coverage for both INR billing flows to assert that no billing data is persisted in that case. The focused unit tests pass locally.

yingbull (approved):

yingbull:
@DragonFSKY  thank you!

Issue #2798, opened 2026-06-11
[Security] JAX-RS REST hasPrivilege sweep — residual: ConsentService (3 endpoints), AppService.getApps, ReportByTemplateService review; the other 15 listed services are now guarded


Ben-Heerema:
## Summary

The majority of JAX-RS REST services under `webserv/rest/` authenticate the caller via `AuthenticationInInterceptor` (session) or `OAuthInterceptor` (OAuth 1.0a), but almost none of them call `securityInfoManager.hasPrivilege()` before executing business logic. Any user who can log in — regardless of their assigned role or security objects — can invoke these endpoints.

## Affected Files

All services reviewed under `src/main/java/io/github/carlos_emr/carlos/webserv/rest/`:

- `ConsentService`, `ReportingService`, `PharmacyService`, `PatientDetailStatusService`, `PersonaService`, `ProgramService`, `AppService`, `DemographicMergeService`, `OscarJobService`, `RecordUxService`, `StatusService`, `ReportByTemplateService`, `ResourceService`, `DiseaseRegistryService`, `FormsService`, `RxLookupService`, `EFormService`, `EFormsService`

## Details

CARLOS's session-based REST services (`/ws/rs/*`) and OAuth REST services (`/ws/services/*`) establish that the caller is a valid, logged-in CARLOS user. However, the base class `AbstractServiceImpl` provides `getLoggedInInfo()` / `getCurrentProvider()` but contains no built-in privilege enforcement. Individual service methods are responsible for calling:

```java
securityInfoManager.hasPrivilege(loggedInInfo, "_secobj", "r", null)
```

This call is absent from the vast majority of REST service methods. In contrast, the `SecurityInfoManager.hasPrivilege()` contract is clearly established in `CLAUDE.md` as required for all resource-access and mutation actions.

## Impact

Any valid CARLOS login (including low-privilege users such as receptionists or read-only accounts) can:
- Read and modify clinical forms (`FormsService`, `EFormService`)
- Invoke disease-registry queries (`DiseaseRegistryService`)
- Run billing and reporting queries (`ReportingService`, `ReportByTemplateService`)
- Submit prescription lookups (`RxLookupService`)
- Manipulate program and patient status (`ProgramService`, `PatientDetailStatusService`)
- Trigger demographic merge operations (`DemographicMergeService`)

## Priority

2-High — Authentication is present; authorization is absent. Exploitable by any authenticated user.

## Fix

[...]

PR #3090, opened 2026-06-29
fix(security): enforce hasPrivilege on ProgramService and PharmacyService JAX-RS endpoints (#2798)

Ben-Heerema:
## Summary

Part of #2798 (does **not** close it — this covers 2 of the remaining 8 services).

The `ProgramService` (`/ws/rs/program`) and `PharmacyService` (`/ws/rs/pharmacies`) JAX-RS services authenticated the caller but performed **no authorization** — any logged-in user (and, per the unauthenticated `/ws/` gap noted in #2798, potentially anonymous callers) could read program/admission data and perform full pharmacy CRUD.

This PR adds the missing `securityInfoManager.hasPrivilege()` checks to every endpoint in both services, the "Domain CRUD" pairing from the issue's scope-tracking comment.

## Scope

- **ProgramService** — guard `/patientList` and `/programList` with `_pmm_management` `"r"` (the security object every PMmodule action uses).
- **PharmacyService** — guard the two GET reads with `_rx` `"r"`, and the `addPharmacy` (POST) / `updatePharmacy` (PUT) / `removePharmacy` (DELETE) mutators with `_rx` `"w"` (matching the existing Rx pharmacy actions `RxManagePharmacy2Action` / `ViewViewPharmacy2Action`).
- Denied callers throw `AccessDeniedException` **before** any business logic runs, consistent with the just-merged in-package `RxWebService` pattern (#3010). Because the guard fails closed when there is no `LoggedInInfo`, this also covers the unauthenticated case.
- **Boundary:** only these two services; no interceptor/auth-path changes, no behavior change for authorized callers, no other #2798 services touched.

## Tests

- `ProgramServiceUnitTest` — allow + deny for `getPatientList` and `getProgramList`; deny path asserts the managers are never invoked.
- `PharmacyServiceUnitTest` — allow path for the list read, plus deny path for all five endpoints, asserting the DAO is never touched.
- Mocks injected via reflection + a testable subclass overriding `getLoggedInInfo()`, mirroring `EFormServiceUnitTest` / `RxWebServiceUnitTest`.

Verified locally:
- `mvn test-compile` — BUILD SUCCESS
- `mvn surefire:test -Dtest='ProgramServiceUnitTest,PharmacyServiceUnitTest'` — Tests run: 10, Failures: 0, Errors: 0

## Summary by Sourcery

Add authorization checks to ProgramService and PharmacyService JAX-RS endpoints to ensure only users with appropriate privileges can access program data and perform pharmacy CRUD operations.

Bug Fixes:
- Restrict ProgramService patient and program list endpoints to callers with _pmm_management read privilege.
- Restrict PharmacyService read endpoints to callers with _rx read privilege and mutator endpoints to callers with _rx write privilege.
- Block unauthenticated or unauthorized access to these endpoints by throwing AccessDeniedException before any business logic executes.

Tests:
- Add ProgramServiceUnitTest to cover allowed and denied access paths for patient and program list endpoints, ensuring managers are not invoked on denied requests.
- Add PharmacyServiceUnitTest to cover allowed and denied access paths for all pharmacy endpoints, ensuring the DAO is not invoked on denied requests.

---
## Summary by cubic
Adds missing authorization to Program and Pharmacy JAX-RS endpoints to block unauthorized reads/writes. Part of #2798; behavior is unchanged for authorized users.

- **Bug Fixes**
  - Program: guard `/patientList` and `/programList` with `_pmm_management` read.
  - Pharmacy: guard GET with `_rx` read; POST/PUT/DELETE with `_rx` write.
  - Unauthorized requests throw `AccessDeniedException` before any business logic, covering unauthenticated callers.
  - Tests: add unit tests for allow/deny paths; assert the correct action is required (`"r"` for reads, `"w"` for writes) and that managers/DAO aren’t called on deny.

<sup>Written for commit dcb0d0a2e7b22a82b9adf684a9fc6dbc4fffd5bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3090?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/main/java/io/github/carlos_emr/carlos/webserv/rest/PharmacyService.java
+++ b/src/main/java/io/github/carlos_emr/carlos/webserv/rest/PharmacyService.java
@@ -39,7 +39,9 @@ import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.MediaType;
 
 import io.github.carlos_emr.carlos.commn.dao.PharmacyInfoDao;
+import io.github.carlos_emr.carlos.commn.exception.AccessDeniedException;
 import io.github.carlos_emr.carlos.commn.model.PharmacyInfo;
+import io.github.carlos_emr.carlos.managers.SecurityInfoManager;
 import io.github.carlos_emr.carlos.webserv.rest.conversion.PharmacyInfoConverter;
 import io.github.carlos_emr.carlos.webserv.rest.to.OscarSearchResponse;
 import io.github.carlos_emr.carlos.webserv.rest.to.model.PharmacyInfoTo1;
@@ -56,8 +58,14 @@ import org.springframework.stereotype.Component;
 @Consumes(MediaType.APPLICATION_JSON)
 public class PharmacyService extends AbstractServiceImpl {
 
+    /** Security object guarding pharmacy management, matching the Rx pharmacy actions. */
+    private static final String SECURITY_OBJECT = "_rx";
+
     @Autowired
     private PharmacyInfoDao pharmacyInfoDao;
+
+    @Autowired
+    private SecurityInfoManager securityInfoManager;
 
     private PharmacyInfoConverter converter = new PharmacyInfoConverter();
 
@@ -73,6 +81,9 @@ public class PharmacyService extends AbstractServiceImpl {
     @GET
     @Path("/")
     public OscarSearchResponse<PharmacyInfoTo1> getPharmacies(@QueryParam("offset") Integer offset, @QueryParam("limit") Integer limit) {
+        if (!securityInfoManager.hasPrivilege(getLoggedInInfo(), SECURITY_OBJECT, "r", null)) {
+            throw new AccessDeniedException(SECURITY_OBJECT, "r");
+        }
         OscarSearchResponse<PharmacyInfoTo1> result = new OscarSearchResponse<PharmacyInfoTo1>();
         result.getContent().addAll(converter.getAllAsTransferObjects(getLoggedInInfo(), pharmacyInfoDao.findAll(offset, limit)));
         return result;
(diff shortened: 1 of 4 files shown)


Ben-Heerema:
@coderabbitai review
@cubic-dev-ai review

Ben-Heerema:
@coderabbitai review
@cubic-dev-ai review

Ben-Heerema:
Thanks @sourcery-ai — both are fair observations. After review, I'm going to leave the code as-is on this PR, with reasoning:

**1. Extracting a `requirePrivilege(String action)` helper.** Legitimate DRY point, but I'm keeping the checks inline:
- It matches the established `webserv/rest` convention — the recently-merged `RxWebService` (#3010) and `EFormService` both inline the `hasPrivilege` check at every endpoint. A helper here would make this PR read differently from its neighbors.
- For authz code specifically, the inline form keeps the security object *and* action visible at each entry point, so an auditor (or a grep) can confirm every method is guarded and with what. That legibility is worth a little duplication.

If we want this consolidated, I'd rather do it as a package-wide cleanup that also touches `RxWebService`/`EFormService` so the convention stays consistent, rather than diverging in one file.

**2. Sharing `SECURITY_OBJECT` between service and test.** I'm intentionally keeping the test's copy separate:
- The production constant is `private`; referencing it would mean widening production visibility purely for test convenience.
- More importantly, the duplication is load-bearing. The tests stub `hasPrivilege(..., eq(SECURITY_OBJECT), ...)`, so if the production security object ever changes without the test being updated, the matcher stops matching the real call → the allow-path test fails. That independent copy is deliberately a tripwire for an unintended secobj change; importing the shared constant would make both sides move together and silence it.

So for this case the "silent divergence" risk actually cuts the other way — the divergence is meant to be loud. Appreciate the review!

Ben-Heerema:
@yingbull ready for review

yingbull (approved):

PR #635, opened 2026-03-14
update: migrate Bootstrap 3.0.0 and 5.0.2 to Bootstrap 5.3.3 across project

yingbull:
### **User description**
Replace all Bootstrap 3.0.0 and 5.0.2 CSS/JS includes with the installed
Bootstrap 5.3.3 library. Update Bootstrap 3 CSS classes to Bootstrap 5
equivalents across 187 files:

- CSS/JS includes: 3.0.0 → 5.3.3, 5.0.2 → 5.3.3 (use bundle.min.js)
- Data attributes: data-toggle/dismiss/target → data-bs-* prefix
- Button classes: btn-default → btn-secondary, btn-large → btn-lg
- Grid classes: col-xs-* → col-*, span* → col-md-*, row-fluid → row
- Icons: glyphicon-* → Font Awesome 6 fa-solid equivalents
- Components: panel → card, well → card card-body bg-body-tertiary
- Labels: label-* → badge bg-*
- Badges: badge-* → bg-*
- Utilities: pull-left/right → float-start/end, text-left/right → text-start/end
- Alerts: close → btn-close, alert-dismissable → alert-dismissible
- Form: input-group-addon → input-group-text
- Remove bootstrap-theme.min.css and DT_bootstrap.css references

___

### **Description**
- Migrated the project from Bootstrap 3.0.0 and 5.0.2 to Bootstrap 5.3.3.
- Updated various components and classes to align with Bootstrap 5 standards.
- Enhanced UI elements to improve consistency and modernize the interface.
- Replaced deprecated classes and attributes with their Bootstrap 5 counterparts.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>configureFax.jsp</strong><dd><code>Update Bootstrap References and Classes in configureFax.jsp</code></dd></summary>
<hr>

src/main/webapp/admin/configureFax.jsp
<li>Updated Bootstrap CSS and JS references to version 5.3.3.<br> <li> Changed alert classes from 'alert-error' to 'alert-danger'.<br> <li> Replaced Bootstrap grid classes with Bootstrap 5 equivalents.<br> <li> Updated button and badge classes to use Bootstrap 5 styles.<br>

</details>

  </td>
[...]

Issue #11, opened 2026-02-03
Hibernate 6 Prep: Migrate typed setters (setInteger, setString, etc.) to setParameter()


yingbull:
## Summary
Replace Hibernate-specific typed parameter setters with JPA-standard `setParameter()`.

## Background
Methods like `setInteger()`, `setString()`, `setLong()` are removed in Hibernate 6. The generic `setParameter()` method works in both versions.

## Scope
- **4 occurrences** across **2 files**

## Affected Files
- `DemographicDaoImpl.java` (lines 137, 2972)
- `ProgramClientStatusDAOImpl.java` (lines 94, 95)

## Migration Pattern
```java
// Before (removed in H6)
query.setInteger("id", 123);
query.setString("name", "test");
query.setLong(1, 456L);

// After (works in H5 AND H6)
query.setParameter("id", 123);
query.setParameter("name", "test");
query.setParameter(1, 456L);
```

## Verification
- [ ] All typed setters replaced with `setParameter()`
- [ ] Type inference works correctly
- [ ] Unit/integration tests pass

yingbull:
completed

Issue #10, opened 2026-02-03
Hibernate 6 Prep: Migrate uniqueResult() to getSingleResult()


yingbull:
## Summary
Replace Hibernate-specific `.uniqueResult()` calls with JPA-standard `.getSingleResult()`.

## Background
`Query.uniqueResult()` is deprecated in Hibernate 6. The JPA-standard `getSingleResult()` works in both versions.

## Scope
- **4 occurrences** across **2 files**

## Migration Pattern
```java
// Before
Object result = query.uniqueResult();

// After (works in H5 AND H6)
try {
    Object result = query.getSingleResult();
} catch (NoResultException e) {
    // Handle no result case
}
```

**Note**: `getSingleResult()` throws `NoResultException` if no result found, while `uniqueResult()` returns null. Handle accordingly.

## Affected Files
- `CaseManagementNoteDAOImpl.java`
- `DemographicDaoImpl.java`

## Verification
- [ ] All `.uniqueResult()` calls replaced
- [ ] Null handling converted to exception handling where needed
- [ ] Unit/integration tests pass

yingbull:
completed

PR #3096, opened 2026-06-30
Investigate provider-to-patient email functionality in CARLOS

Ben-Heerema:
## Summary

This PR investigates how provider-to-patient email functionality currently works in CARLOS, including the available user flows, backend routing, configuration requirements, and expected behavior in local development.

## Scope

- Identify all UI entry points for provider-to-patient email.
- Trace the backend actions and services used to compose and send email.
- Confirm required permissions, patient data, consent, and sender configuration.
- Verify whether email is available as a direct chart workflow, eForm workflow, or both.
- Document local development behavior and any expected setup requirements.
- Capture gaps, confusing UX, or failure modes for follow-up work.

## Validation Plan

- Test the provider workflow from a patient chart.
- Test the eForm email workflow.
- Confirm behavior when the patient has no email address.
- Confirm behavior when consent is missing or not explicit opt-in.
- Confirm behavior when no sender account is configured.
- Confirm behavior when SMTP/API delivery is unavailable.
- Review email logs/status handling after attempted sends.

## Notes

Detailed findings will be added as PR comments as each part of the workflow is traced and verified.

## Summary by Sourcery

Documentation:
- Add a specification document outlining the purpose, scope, and validation plan for investigating provider-to-patient email workflows in CARLOS.

---
## Summary by cubic
Adds a documentation spec mapping provider‑to‑patient email in CARLOS: UI entry points, backend routes, required permissions/consent, sender config, and local setup. Includes a validation plan to test chart/eForm flows, edge cases (no email/consent, delivery outages), and review logs/status to capture gaps and failure modes.

<sup>Written for commit a5cbedc4327a095b0031af2bd7ad06dae3a34385. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/specs/provider-patient-email-investigation.md
+++ b/docs/specs/provider-patient-email-investigation.md
@@ -1,0 +1,28 @@
+# Provider-to-Patient Email Investigation
+
+## Purpose
+
+Investigate how provider-to-patient email functionality works in CARLOS, including available user flows, backend routing, configuration requirements, and expected behavior in local development.
+
+## Scope
+
+- Identify UI entry points for provider-to-patient email.
+- Trace backend actions and services used to compose and send email.
+- Confirm required permissions, patient data, consent, and sender configuration.
+- Verify whether email is available as a direct chart workflow, eForm workflow, or both.
+- Document local development behavior and expected setup requirements.
+- Capture gaps, confusing user experience, and failure modes for follow-up work.
+
+## Validation Plan
+
+- Test provider workflow from a patient chart.
+- Test eForm email workflow.
+- Confirm behavior when the patient has no email address.
+- Confirm behavior when consent is missing or not explicit opt-in.
+- Confirm behavior when no sender account is configured.
+- Confirm behavior when SMTP or API delivery is unavailable.
+- Review email logs and status handling after attempted sends.
+
+## Findings
+
+Detailed findings will be added as pull request comments as the workflow is traced and verified.


Ben-Heerema:
### Finding: active provider-to-patient email path

The provider-to-patient send path currently appears to be the eForm email workflow:

1. Provider opens a patient eForm.
2. Provider clicks **Email** in the floating eForm toolbar.
3. The toolbar sets `emailEForm=true` and saves the eForm.
4. `AddEForm2Action` stores the email context in session and redirects to `/email/emailComposeAction`.
5. `EmailCompose2Action` prepares the compose screen with patient recipient, consent status, sender accounts, attachments, encryption settings, and chart-note options.
6. `EmailSend2Action` sends using the selected configured sender.

Relevant files:
- `src/main/webapp/eform/eformFloatingToolbar/eform_floating_toolbar.js`
- `src/main/java/io/github/carlos_emr/carlos/eform/actions/AddEForm2Action.java`
- `src/main/java/io/github/carlos_emr/carlos/email/action/EmailCompose2Action.java`
- `src/main/java/io/github/carlos_emr/carlos/email/action/EmailSend2Action.java`

Ben-Heerema:
### Finding: CARLOS Messenger is separate from patient email

The internal Messenger UI can associate a demographic/patient with a message, but its recipient model is provider/group based. I did not find evidence that the Messenger compose flow directly sends patient-facing email.

The patient-facing email route found so far is the eForm email workflow, not `/messenger/CreateMessage`.

Relevant files:
- `src/main/webapp/WEB-INF/jsp/messenger/CreateMessage.jsp`
- `src/main/webapp/WEB-INF/classes/struts-messenger.xml`
- `src/main/java/io/github/carlos_emr/carlos/managers/MessengerDemographicManagerImpl.java`

Ben-Heerema:
### Finding: email availability depends on several gates

The eForm toolbar Email button and compose flow depend on these conditions:

- Current user must have `_email` privilege.
- Email consent tracking must be configured.
- At least one active sender account must exist.
- The patient demographic record must include at least one valid email address.
- Consent status is shown in the compose screen; if the patient is not `Explicit Opt-In`, the eForm toolbar prompts the provider to type `Yes` before continuing.

If the email system is not configured, the toolbar disables the Email button with the title `Consent or Sender Details Not Configured`. If the user lacks `_email`, the toolbar hides the button.

Relevant files:
- `src/main/webapp/WEB-INF/jsp/eform/eformFloatingToolbar/eform_floating_toolbar.jspf`
- `src/main/webapp/eform/eformFloatingToolbar/eform_floating_toolbar.js`
- `src/main/java/io/github/carlos_emr/carlos/managers/EmailComposeManager.java`

Ben-Heerema:
### Finding: local development SMTP failure is expected without a mail service

In the current local develop runtime, email delivery reached the SMTP send step but failed because the configured sender tried to connect to `localhost:25` and no SMTP server is listening there.

Observed error:

```text
Mail server connection failed. Failed messages: org.eclipse.angus.mail.util.MailConnectException: Couldn't connect to host, port: localhost, 25; timeout -1; nested exception is: java.net.ConnectException: Connection refused
```

Runtime/default properties point email at local SMTP:

```properties
email.host=localhost
email.port=25
```

This confirms the compose/send path is active, but local delivery requires either:

- a local SMTP/mail-catcher service listening on the configured host/port,
- a real SMTP sender configuration,
- or an API sender configuration such as SendGrid.

Relevant files/config:
- `/root/carlos.properties`
- `src/main/resources/carlos.properties`
- `src/main/java/io/github/carlos_emr/carlos/email/helpers/SMTPEmailSender.java`
- `src/main/java/io/github/carlos_emr/carlos/email/core/EmailSender.java`

Ben-Heerema:
### Finding: no general chart-level email composer found yet

I found the eForm-launched patient email workflow and the admin resend workflow, but I did not find a general provider chart action that opens a blank patient email composer directly from the chart without eForm context.

This may be a product gap or a workflow decision worth confirming:

- If the intended behavior is only `Email from eForm`, the UI/docs should make that clear.
- If providers should be able to send a plain message to a patient from the chart, that appears to need a separate direct-entry flow.
- The existing email compose JSP already has a `DIRECT` transaction mode, but the normal patient-facing entry point found so far is eForm-based.

Relevant files:
- `src/main/webapp/WEB-INF/jsp/email/emailCompose.jsp`
- `src/main/java/io/github/carlos_emr/carlos/email/action/EmailSend2Action.java`
- `src/main/java/io/github/carlos_emr/carlos/email/admin/ManageEmails2Action.java`

Ben-Heerema:
Current general steps to access email screen:

1. Login as usual
2. Open patient search (I used FAKE-FOREMAN)
3. Open patients chart
4. Open EChart
5. Click the "+" on eForm
6. Click on Rich Text Letter (an old eForm)
It should open the following, which may also need a UI update:
image

7. Click Email in the bottom bar 
8. It will not have consent set up for FAKE-FOREMAN, so just type "yes" to get past consent. 
It will open the following which needs some UI updates:
image

Ben-Heerema:
Current thoughts:
Add a warning that subject is not encrypted
Email UI seems to be broken/needs touch ups to make usability better. 
While encryption is correctly turned on by default there should be at least one or two warnings when you turn it off.
This takes 5 pop-ups to access, should it be moved? into msg?
Change config for postfix so it writes to a file and give instructions to enable in dev container

Ben-Heerema:
### Finding: email compose UI layout issue is existing `develop` behavior

The weird email compose layout in the screenshot does **not** come from this investigation PR. The PR diff only adds `docs/specs/provider-patient-email-investigation.md`; it does not modify the email UI.

The immediate compose-screen cause is in `emailCompose.jsp`: the top-level form is rendered as a Bootstrap flex container:

```jsp
<form id="emailComposeForm" class="d-flex flex-wrap align-items-center gap-2" ...>
```

All of the major sections (`From`, `To`, `Subject`, `Body`, `Encryption`, `Additional options`, `Attachments`) are direct child `.card` elements with no explicit width. In Bootstrap flex layout, those cards become content-sized flex items and wrap wherever they fit. That produces the scattered/masonry-like layout shown in the screenshot: `From` and `To` on one row, `Subject` beside `Body`, `Encryption` floating to the right, and the lower cards wrapping independently.

This appears to be fallout from the Bootstrap migration commit `c83b95c2bcc` (`update: migrate Bootstrap 3.0.0 and 5.0.2 to Bootstrap 5.3.3 across project (#635)`). Before that, the form used `class="form-inline"`; under Bootstrap 5 that class is effectively gone, but the form still behaved like a normal block element, so block-level cards would naturally stack/stretch. The migration replaced it with `d-flex flex-wrap align-items-center gap-2`, which changed the layout model for the entire compose form.

Suggested direction:

- Remove the flex layout from `#emailComposeForm` and make it a normal vertical compose form, or wrap cards in an explicit Bootstrap grid.
- Give top-level sections intentional width, for example `w-100` for full-width rows or `col-*` inside a `row g-3` if some sections should sit side-by-side.
- Rework narrow internal label columns like `col-sm-1` in the `To` and encryption cards. They are too narrow for labels such as `Email(s)` and `Encrypt Attachments`, which is why labels visually collide/truncate in the screenshot.
- Move the encryption switch out of the `<h5>` text flow into a flex card header, e.g. header title on the left and switch on the right.
- Consider adding explicit warning copy for unencrypted fields: the UI currently warns that the body is unencrypted, but the subject line is also normal email metadata and should be treated as unencrypted.

### Separate issue: Rich Text Letter screenshot shows injected logout script as content

The first screenshot also shows a raw `<script>(function(){...SessionHeartbeat...})()</script>` block inside the Rich Text Letter editor content. That looks like `LogoutBroadcastFilter` output leaking into the editable eForm document.

Relevant path:

- `efmformadd_data.jsp` prints `thisEForm.getFormHtml()` directly as the eForm HTML response.
- `LogoutBroadcastFilter` is mapped broadly to `/*` for `REQUEST` and `FORWARD` and currently only excludes `/logoutPage` and `/status/SessionHeartbeat`.
- The filter appends the session heartbeat/logout script to authenticated `text/html` responses.
- Rich Text Letter uses an iframe/editor that reads and writes document body content through `innerHTML`/`textContent`. If the filter appends application chrome script into that eForm document, the editor can treat it as letter content instead of page chrome.

This should be tracked separately from the email compose layout fix. Likely fixes are either adding targeted exclusions for eForm content/template document routes, or changing the injection strategy so the logout script is only added to app shell pages where it cannot become provider-authored document content.

Relevant files:

- `src/main/webapp/WEB-INF/jsp/email/emailCompose.jsp`
- `src/main/webapp/WEB-INF/web.xml`
- `src/main/java/io/github/carlos_emr/carlos/app/LogoutBroadcastFilter.java`
- `src/main/webapp/WEB-INF/jsp/eform/efmformadd_data.jsp`
- `src/main/webapp/WEB-INF/eform-assets/editControl2.js`

Ben-Heerema:
## Production-readiness assessment: provider-to-patient email

Following up on the investigation findings above, here is a code-level assessment of what stands between the current email feature and production use with real patient data. This is based on reading the actual send path (`EmailCompose2Action` → `EmailSend2Action` → `EmailManager` → `EmailSender` and its SMTP/SendGrid helpers), not just the UI behavior.

**TL;DR:** the plumbing works and privilege gating is solid (`_email` checks are layered through actions, managers, and senders), but there are five hard blockers — mostly PIPEDA/PHI compliance — plus operational gaps around configuration, reliability, and cleanup. None of the currently in-flight PRs (#3097, #3100, #3101) address the blockers.

---

### Blockers (must fix before any real patient email)

**1. Consent is displayed but never enforced.**
`EmailComposeManager.getEmailConsentStatus()` (lines 292–315) resolves the patient's `ConsentType` via `PatientConsentManager`, but the result is only rendered as a warning banner in `emailCompose.jsp` and echoed back as a hidden field. Nothing on the server blocks a send when the status is `Explicit Opt-Out` or `Unknown`:

- `EmailSend2Action.sendEmail()` / `prepareEmailFields()` (lines 196–265) never read consent.
- `EmailManager.sendEmail()` (lines 125–147) never checks consent.
- The client-side `validateForm()` in `emailCompose.jsp` (lines 691–731) validates sender/recipient/subject/body only — no consent check — and the `isEmailAutoSend` path (lines 790–795) can auto-submit the form on page load with no consent gate at all.
- The consent status at time of send is **not recorded on the `EmailLog`**, so a send to an opted-out patient succeeds and leaves no audit trail of the override.

This is the single biggest compliance gap. Enforcement belongs in `EmailManager.sendEmail()` (server-side, not JS), with the consent status persisted to the log and an explicit, audited override path if the product wants one.

**2. The "encryption" scheme is weak and self-defeating.**
The encryption toggle does not encrypt the email. It password-protects PDF attachments via PDFBox (`EmailManager.encryptEmail()`, lines 452–472; `PDFEncryptionUtil`). The problems:

- The password is `DOB(YYYYMMDD) + HIN` (`EmailComposeManager.createEmailPDFPassword()`, lines 444–447) — low-entropy and derived entirely from PHI.
- The cleartext email body includes a clue that states that exact recipe: "your date of birth in the format YYYYMMDD followed by the 10 digits of your health insurance number" (`EmailCompose2Action.java:246`). Anyone who intercepts the email is told how to construct the password.
- The message body itself is never encrypted — transport TLS only. There is no S/MIME or PGP.
- The password is stored **in plaintext** in `emailLog.password` (`EmailLog.java:129`), surfaced in the Manage Emails admin UI (`EmailStatusResult.getPassword()`), and written into the chart note by `EmailNoteUtil` (lines 216–285).

This needs a design decision, not a patch: random per-message passwords delivered out-of-band (SMS/phone), a portal-link model, or dropping the "encrypted" claim from the UI entirely. PR #3100's warnings are honest UX but don't change the underlying scheme.

**3. `EmailSend2Action` accepts GET.**
`execute()` routes purely on the `method` request *parameter* with no HTTP-verb check (`EmailSend2Action.java:79–91`), so a crafted GET can send an email and persist state — CSRF exposure. This violates the repo's mutator GET/HEAD-rejection contract; the fix should register the action in `MutatorActionGetRejectionContractTest` alongside the verb check.

**4. Credentials and secrets stored in plaintext.**
SMTP passwords and SendGrid API keys live unencrypted in the `emailConfig.configDetails` TEXT column (`EmailConfig.java:42–43`). Combined with the plaintext PDF passwords in `emailLog` (see #2), a database read exposes both transport credentials and every "encrypted" attachment's password.

**5. PHI temp files are never cleaned up.**
Every compose/send/resend renders patient PDFs to the temp directory (`ConvertToEdoc.saveAsTempPDF`, encrypted copies as `encryptedPDF_<millis>.pdf` via `PDFEncryptionUtil.java:22`), and nothing ever deletes them — no `Files.delete`, no `deleteOnExit`, no cleanup job. Unencrypted patient PDFs accumulate on disk indefinitely. (The only deletion in the flow is `removeEFormData` removing the eForm DB record — unrelated to temp files.)

---

### Operational gaps

**6. No admin UI for sender accounts.** `configureEmail.jsp` (behind `ViewConfigureEmail2Action`) is a static documentation page with sample JSON; `EmailConfig` rows must be inserted into the DB by hand, with no validation until send time. A null/missing sender config throws an `IllegalArgumentException` from `prepareEmailForOutbox()` (`EmailManager.java:173–179`) that `sendEmail()`'s `catch (EmailSendingException)` does **not** catch — so misconfiguration surfaces as an HTTP 500 rather than a graceful failed-status log.

**7. Synchronous, fire-once sending.** `EmailManager.sendEmail()` runs the full network send in the HTTP request thread. No queue, no async, no retry, no rate limiting. A slow SMTP server blocks Tomcat workers; failures are terminal until an admin manually resends from Manage Emails.

**8. No attachment size limits.** No per-attachment or total-message cap before the SMTP send or the SendGrid Base64 encode — oversized messages just fail at the provider (SendGrid ~30 MB) with a runtime error.

**9. Deliverability/ops prerequisites (outside the code).** A production sender domain with SPF/DKIM/DMARC, a real relay or SendGrid account, and monitoring of FAILED rows in the email log. PR #3097's Postfix capture covers the devcontainer only.

---

### Hardening (important, not blocking)

**10. Test coverage on the core path is nearly empty.** `EmailSend2ActionTest` has a single test (the `cancel()` redirect). There are no tests for `EmailManager`, `EmailComposeManager`, `EmailSender`, either SMTP sender, `APISendGridEmailSender`, or `EmailNoteUtil`, and no Playwright email flow in `scripts/` or `package.json`.

**11. Assorted send-path issues:**
- `EmailAttachmentSettings.sanitizeSubject()` strips CRLF on the eForm compose entry path, but the actual send path reads `subjectEmail`/recipients straight from `request.getParameter` with no sanitization (`EmailSend2Action.java:224–238`). JavaMail provides some protection; it should not be the only layer.
- Recipient email is emitted unencoded in `emailCompose.jsp:298` and `:302` (`value="${receiverEmail}"`) — low risk since the values are RFC-validated from the demographic record, but inconsistent with the null-safe encoding used everywhere else on that page.
- `APISendGridEmailSender` sets `setNeedClientAuth(true)` on a throwaway copy of default SSL parameters (lines 166–168) — a misapplied server-side mTLS flag that does nothing — and redundantly embeds the API key in the JSON request body (lines 267–273).
- `EmailManager.java:144` logs the raw provider exception; SMTP/SendGrid error text can echo recipient addresses into log4j — a plausible PHI leak channel.

---

### Positive findings, for balance

- `_email` privilege checks are consistently applied at action, manager, and sender layers, with per-type checks (`_eform`/`_edoc`/`_lab`/`_hrm`/`_form`) on attachment preparation.
- External SMTP enforces STARTTLS (required, TLSv1.2) rather than making it optional.
- `PathValidationUtils` is used correctly for attachment file resolution and secure temp-file creation.
- Attachment filenames are sanitized to generic `attachment_NNN.pdf`, keeping PHI out of filenames.
- `EmailLog` + `LogAction` auditing and the chart-note integration (`EmailNoteUtil`) are a solid foundation.

### Suggested sequencing

1. Merge the three in-flight PRs (#3097, #3100, #3101) — green, review-blocked only.
2. Fix consent enforcement (#1), GET rejection (#3), and temp-file cleanup (#5) — well-scoped, established codebase patterns exist for all three.
3. Decide the encryption/password design (#2) and secrets-at-rest approach (#4) — these need product/security input before code.
4. Sender-account admin UI (#6) and async/retry sending (#7).
5. Test coverage (#10) and the hardening list (#11), then deliverability ops (#9).

Also still open from this investigation: whether a direct chart-level "email patient" entry point (outside the eForm flow) is a product requirement — that remains a question for discussion, not a defect.

---
*Assessment generated with [Claude Code](https://claude.com/claude-code); file/line references verified against `develop` as of 2026-07-02.*

Ben-Heerema:
> **Merge-order update (2026-08-14):** the ordering in this historical rebase note is superseded by the [canonical merge train](https://github.com/carlos-emr/carlos/pull/3096#issuecomment-5294481647). The rebase and branch-analysis details below remain useful historical context.

## Email PR stack: merge order, current state, and one hard blocker

The ordering for this series was documented across seven different PR bodies and contradicted itself in places (#3203 said "merge ordering needs to be settled"). This comment consolidates it in one place and records the current state after a rebase pass.

### Review / merge order

```
  #3116  GET/HEAD rejection ─────────────────┐
  #3128  consent enforcement ────────────────┤
  #3188  sender misconfig handling ──────────┼──  independent, land in any order
  #3187  operations runbook (docs) ──────────┤
  #3097  devcontainer mail capture ──────────┘

  #3127  merge body + encrypted message field    (independent)

  #3130  secrets at rest  ──►  #3135  random passphrases  ──►  #3218  PDF signing

  #3212  InnoDB pre-migration
     └─► #3138  archive foundation
            ├─► #3202  SMTP archive
            ├─► #3203  SendGrid archive
            └─► #3204  archive access hardening
                   └─► #3222  single eligibility predicate + read entry point
```

### State after this pass

**Ready for review (9)** — all merge cleanly against `develop`:

| PR | Notes |
|----|-------|
| #3116 | rebased; 49 tests pass incl. the mutator GET-rejection contract |
| #3127 | rebased (import-only conflict) |
| #3128 | already clean, untouched |
| #3130 | rebased; non-trivial merge, see note below |
| #3135 | rebased; 66/66 email tests pass |
| #3187, #3188, #3097 | already clean, untouched |
| #3202 | **re-based onto #3138 instead of `develop`** — diff dropped from 22 commits to 13 files, and it went from CONFLICTING to MERGEABLE |

**Converted to draft (4)** — cannot be merged as-is, so they no longer sit in the review queue pretending to be actionable: #3138, #3203, #3204, #3218.

### Hard blocker: the archive chain predates the Flyway migration

`develop` moved to Flyway in #3150 ("consolidated V1 baseline, legacy build retired") and **deleted `database/mysql/oscarinit.sql`**. `database/mysql/migration/README.md` now describes itself as the single source of truth for the schema.

Every PR in the archive chain still ships its schema the old way — appending to `oscarinit.sql` plus a dated `database/mysql/updates/update-2026-07-*.sql`:

| PR | Schema files |
|----|--------------|
| #3138 | `oscarinit.sql`, `update-2026-07-07-outbound-email-archive.sql` |
| #3202 | same two |
| #3203 | those plus `update-2026-07-06-outbound-email-archive-reference-engines.sql` |
| #3204 | same two |
| #3212 | `update-2026-07-06-outbound-email-archive-reference-engines.sql` |

Merging develop into #3138 produces a `modify/delete` conflict on a file that no longer exists, and the dated `update-*.sql` scripts are no longer in the execution path at all. **This is a schema-authoring rewrite into versioned Flyway migrations, not a merge conflict** — the three archive tables, their foreign keys, and the InnoDB pre-check all need to be re-expressed as `V1.0.x__*.sql` files under `migration/common/`. It needs a decision on version numbering against whatever else is queued for the next Flyway version, so it should not be guessed at.

#3212 currently reports MERGEABLE, but only because nothing it touches was deleted — its migration is equally obsolete and would silently never run.

Additionally, #3203 and #3204 have diverged from each other and from #3138; no choice of base branch makes either merge cleanly (tested against `develop`, #3138, #3202, and #3203). Both carry the full foundation commit set inline. They are best rebuilt as thin deltas on top of #3138 once the Flyway rewrite lands.

### Two things worth a reviewer's attention

**#3130 merge was not mechanical.** `develop` independently refactored `APISendGridEmailSender` (extracting `assertAccepted` and `validateEndpoint`, making `createEmailJSON` private) while #3130 widened those members for testing and added `{ENC}` decryption at send time. Both branches had also *independently created* `APISendGridEmailSenderUnitTest` with disjoint test sets. The resolution keeps develop's transport-acceptance methods, keeps #3130's credential hardening and package-private visibility, and merges both test sets into one class (21 tests, all passing).

**Pre-existing behaviour carried through unchanged, flagged not fixed:** `EmailCompose2Action` appends `e.getMessage()` from a caught `PDFGenerationException | RuntimeException` into a browser-visible error string. That came from #3193 and is untouched here, but it is an information-disclosure smell on a PHI-adjacent path and probably deserves its own ticket.

### Scope cleanup still outstanding

#3204 carries changes unrelated to archive access hardening — eForm POST enforcement (405), OpenPDF 3.0.3 vendoring, a license-URL fix, and a CI MariaDB init-wait bump. The OpenPDF vendoring has since landed on `develop` independently (#3192), so some of that will evaporate on rebase; the rest should be split into a separate chore PR that can merge on its own.

---
*Rebase pass and status audit performed with [Claude Code](https://claude.com/claude-code); branch states verified against `develop` as of 2026-08-13.*

Ben-Heerema:
Recording the decision on the open question from the [investigation findings](https://github.com/carlos-emr/carlos/pull/3096#issuecomment-4844060265): whether a **direct chart-level "email patient" entry point** should exist outside the eForm flow.

**Decision: deferred pending outside input.** The current intent is to add one, but we want external/product feedback before designing it, so it is not being scoped or built now. Recording it here so it stops being an unanswered question buried in a July comment thread.

Ben-Heerema:
> **Protocol update (2026-08-14):** the ownership decisions below still stand, but the original publish-before-transport sequence is superseded by the pending-record protocol in #3449 and the CARLOS implementation ticket #3450.

Two design decisions recorded for the email workstream:

**Passphrase escrow (#3449 / #3450).** The unlock-secret store lives in the **portal**, not CARLOS. #3135 non-persistence stands unchanged. CARLOS stages its generated passphrase as a pending portal record, sends the email, then activates on transport success or revokes on failure; activation failures require durable reconciliation. Portal staging failure blocks by default with an audited, non-disableable break-glass modal. The portal receiver must first support accepting CARLOS-generated passphrases.

**Archive retention (#3448).** Indefinite. Outbound patient email is a legal record of what was sent. Two riders: the controlled-deletion/legal-hold path from #3138 stays for court orders, privacy-law erasure, or misdirected sends, and permanent retention of full RFC822 artifacts including attachments requires an explicit capacity/tiering plan.

The portal and archive remain separate lanes with separate datastores and key domains. No portal work belongs in the core email merge train.

Ben-Heerema:
## Canonical email merge train and follow-on lanes (updated 2026-08-14)

This supersedes older merge-order instructions scattered across the email PR comments. The order below is chosen to minimize conflicts; it does not imply every adjacent PR has a functional dependency.

### Core merge train

```
#3097 dev mail capture
  → #3116 GET/HEAD rejection
  → #3127 single message field
  → #3128 consent enforcement/audit (convert its schema change to Flyway first)
  → #3130 transport-secret hardening
  → #3188 sender/resend/status hardening
  → #3135 random passphrases + server-side compose state
```

Merge one PR, then rebase only the next overlapping PR onto current `develop`. Do not repeatedly merge `develop` into every downstream branch.

#3187 is documentation and should merge after the core behavior/specification updates so the runbook describes what actually landed.

### After #3135

- Implement #3114 temp-file lifecycle cleanup from a fresh `develop` branch.
- Rebuild #3218 as a thin PDF-signing-only delta on top of the landed #3135 behavior. The current draft should not be merged as-is.

### Archive lane — separately owned

#3212 / #3138 / #3202 / #3203 / #3204 are being handled as a separate archive-stack rebuild. The core train above must not wait for or modify those branches. #3448 application-level archive encryption starts only after that archive chain lands.

### Portal lane

#3220 remains a separate draft and is not part of this merge train. The ownership and failure-handling decision is recorded in #3449, and CARLOS implementation is tracked in #3450; no portal code changes belong in the core email PRs.

### Product decisions already settled

- #3135 ships with one-time provider display/manual out-of-band delivery as the interim behavior; it does not wait for SMS or portal integration.
- English-only passphrases are accepted for the transitional MVP.
- A lost passphrase requires resend with a fresh passphrase until #3450 and a compatible portal receiver are deployed.
- Direct chart-level “email patient” remains deferred pending external/product input.

Ben-Heerema:
## Email merge-train progress update — 2026-08-14

The canonical order above remains unchanged:

`#3097 → #3116 → #3127 → #3128 → #3130 → #3188 → #3135`

Work completed while merge approval is pending:

- #3128 obsolete dated schema update is now Flyway `V1.0.9`; schema verification passes, its focused email tests pass, and a live opted-out send is blocked server-side.
- #3188 was live-tested through the stale-selected-sender race and returns a graceful failed-send page rather than a 500.
- #3135 was live-tested through two compose attempts: the random values are distinct and correctly formatted, browser display is readonly/non-authoritative, and the server-side token plus separate-delivery instruction are present.
- #3135 shutdown lifecycle issue is fixed in `9150dc7900`: the focused lifecycle suites pass and a live Playwright-created compose state survived through a clean Tomcat stop with no email cache shutdown failure or leaked email pruner thread.

Screenshots and exact results are attached to the respective PR comments. The archive branches and portal were left untouched. The one current #3128 CI failure is the unrelated repository-wide Surefire naming guard for `DemographicExportAction42ActionRequestMethodTest`, not a file in #3128.

The next merge-side action is still: merge one PR in order, then rebase only the next overlapping PR onto the new `develop`.

Ben-Heerema:
## Email status update: #3114 is now implemented on #3135

The merge plan no longer needs a separate post-#3135 temp-cleanup PR. Per the latest direction, #3135 now owns both:

1. server-generated PDF passphrases / one-time compose state (#3134), and
2. the generated email PDF lifecycle (#3114).

The lifecycle implementation is in `ecad764fe3`: private per-compose directories, immediate ownership of renderer/encryption output, cleanup on every terminal state, active-lease-safe orphan sweeping, source-document preservation, and PHI-free metrics. Focused tests and packaging pass; live Playwright verified cancel and shutdown cleanup. Details and screenshot: https://github.com/carlos-emr/carlos/pull/3135#issuecomment-5296920633

Merge-order impact: keep #3135 in its existing position, but remove any separate #3114 cleanup item after it. Archive work remains a separate lane and was not pulled into this PR.

Ben-Heerema:
## Email workstream status — 2026-08-19

The [canonical merge train](https://github.com/carlos-emr/carlos/pull/3096#issuecomment-5294481647) order is unchanged. This updates state, records three defects found along the way, and revises the archive-lane plan.

### Core train is green and current

`#3097 → #3116 → #3127 → #3128 → #3130 → #3188 → #3135`, then `#3187`.

All seven plus #3187 and #3212 now pass `tests`. Every branch that was behind has been merged forward; #3097, #3128, #3187 and #3188 are at zero commits behind `develop`, and each PR's diff is byte-for-byte what it was before the merge. Nothing is blocked on code. The remaining gate is review approval.

### Why every email PR was showing a red `tests` job

Not seven problems — one stale file. `develop` split `DemographicExportAction42ActionRequestMethodTest` into a shared `DemographicExportActionUnitTestBase` plus per-scenario `*UnitTest` classes in #3410. Every email branch still carried the old file, so `SurefireIncludeCoverageUnitTest` failed on all of them. CI runs against the **branch head**, not the merge ref, so a merge with `develop` would not have cleared it.

Deleted on all 11 email branches. Note for whoever merges the rest: this produces a rename/delete conflict against `develop`, resolved by taking `develop`'s `DemographicExportActionUnitTestBase` and keeping the deletion.

### Flyway versions collided again

Three migrations claimed versions `develop` had since taken. Renumbered in merge order, starting at `V1.0.14` because the pending 2026.08 forward-merge #3473 claims common `V1.0.13` and Ontario `V1.0.12`:

| PR | was | now |
|----|-----|-----|
| #3128 consent audit | `V1.0.9` | `V1.0.14` |
| #3212 archive engine pre-check | `V1.0.9` | `V1.0.15` |
| #3138 archive tables | `V1.0.10` | `V1.0.16` |

#3138's DB schema verify passed after the rename. The migration READMEs still need one "next free version" pass at merge time; that was left out per-branch on purpose, since three branches editing the same line would conflict with each other and with #3473.

**Process point worth adopting:** migration versions should be assigned at merge time, not authoring time. This is the second time this series has collided.

### Three defects found, all from tests that never ran

1. **#3188 shipped two dark tests.** `ManageEmails2ActionTest` and `EmailDataTest` are named plain `*Test`, which Surefire's `<includes>` never select — 157 lines that had never executed. Renamed to `*UnitTest`. Running them for the first time exposed a real break: #3193 added a `SpringUtils`-wired `PdfPreviewCapabilityService` to `ManageEmails2Action`, and the test never registered a mock, so constructing the action threw.
2. **`EmailAttachmentSettingsTest` was wrong and silently absent.** Its max-length test builds a 253-character address and asserts it is 254 — it could never have passed. Fixed and un-darkened on `fix/email-attachment-settings-dark-test`. No production change; the 254-character bound was correct.
3. **`CombinePDF2ActionTest` was dark too** — authorization tests that read like coverage but never ran. Renamed as part of the split below.

### Archive lane — plan revised

The earlier note said #3203 and #3204 should be rebuilt as thin deltas. That is confirmed for #3203 and **wrong for #3204's PDF work**, which turns out not to be cleanly separable.

- **#3202** merged its base back in. It had stopped being a descendant of #3138, missing 11 commits, so its diff-against-base was misleading.
- **#3203 rebuilt** on `feature/3114-sendgrid-outbound-email-archive-rebuild`, on top of #3202. Its 31 files were 28 copies of the foundation plus three real ones. Its `APISendGridEmailSender` was deliberately **not** ported: it predates the validated-endpoint hardening (DNS pinning, redirect rejection, bounded timeouts), so taking it wholesale would have regressed security. The archive capability was written onto the current hardened file instead.
- **#3204's PDF work split out** to `fix/combine-pdf-patient-authorization`, off `develop`. `CombinePDF2Action` had no per-patient authorization: `_edoc` write is module-level, so any user reaching the route could name arbitrary document numbers and get them merged into one PDF. The `CtlDocumentDao` changes are archive-free and lifted cleanly; the action itself also refuses documents backing an archive artifact, which needs `OutboundEmailArchiveDao` and **stays with #3204**. So the split is 4 of 5 files clean, 1 file requiring surgery — not the clean 5-commit lift previously assumed.

### #3222 resolved by construction, not by reconciliation

#3202 and #3203 defined `supportsOutboundArchive()` two mutually exclusive ways, and the plan was for whoever merged second to unify them. The rebuild removes the predicate instead.

`supportsOutboundArchive()` restated, in a second independently maintained place, a decision `EmailSender`'s transport switch already makes. Two statements of one fact can disagree, and here disagreement is silent — the send succeeds and simply is not recorded. Archiving is now a property of the transport: `OutboundEmailTransport` requires each implementation to produce its own artifact bytes, attachment metadata, content type, artifact type and file name, and one switch serves both `send()` and `prepareOutboundArchive()`. A configuration that resolves to no transport throws exactly as before.

On the open question of `EmailType.API` with a non-SendGrid provider: both switches already ended in `default: throw`, so such a config could never send in the first place. Failing closed is not a new restriction, just the existing one stated once instead of twice.

Coverage matches #3222's acceptance criteria: a sweep over the whole `EmailType × EmailProvider` product asserting sending and archiving reach the same verdict for every configuration. It sweeps the enum product rather than today's providers because adding a provider constant with no transport is precisely how a silent gap would appear.

### Open items

- **Review approval on the core train** — the only thing on the critical path.
- **New PR for the SendGrid rebuild** (base `feature/3114-smtp-outbound-email-archive`), then close #3203. Different history, so it cannot be pushed to the existing branch.
- **#3204 proper** — the archive read API and `OutboundEmailArchiveDocumentGuard` wiring, minus the split-out PDF work. Note that #3138 has independently absorbed some of the same hardening, so the remaining delta needs scoping against it rather than assuming the original diff.
- **Neither #3138 nor #3202 runs the `tests` job at all**, because their PR base is a feature branch rather than `develop`. The archive lane has never had test-suite CI and will not until #3212 lands and #3138 retargets `develop`.
- Migration README pass at merge time; one dead baseline line on `develop` from #3389 resurrecting what #3410 deleted (inert, but wrong).

---
*Status pass performed with [Claude Code](https://claude.com/claude-code); branch and CI states verified against `develop` as of 2026-08-19.*

yingbull:
@Ben-Heerema what do we want to do with these assets? Is this material to add to an issue ticket that exists, or do we want stuff in docs, or?  If you could clean up these investigation type tickets appropriately (if it makes sense to carry in the project docs, we can, else an issue ticket with the info in makes sense? or another idea if you have it?).

Ben-Heerema:
@yingbull This was meant to function sort of like an umbrella tracking ticket and should have been a ticket not a PR. The asset is fine to be tossed, this PR just helped me keep what had been/ hadn't been done sorted. Happy to close it and make ticket/s, will add to my TO DO list

yingbull:
Got it. closing.

Issue #2969, opened 2026-06-20
Startup persists a generated encryption key to user.home even when config loaded from WEB-INF — next boot loads the stub and skips WEB-INF config


yingbull:
## Summary

When CARLOS generates a new encryption key at startup, it always persists it to the **user-home** properties file, even if the configuration was actually loaded from `/WEB-INF/`. On the *next* boot the now-existing user-home stub file (which contains only the key) is loaded first, the `if (p.isEmpty())` guard is skipped, and the `/WEB-INF/` configuration — DB credentials and everything else — is **never read**.

`propFileName` is fixed to the user-home path and never reassigned to reflect where config was actually loaded from:

```java
// src/main/java/io/github/carlos_emr/carlos/login/Startup.java  (develop ~line 90)
String propName = contextPath + ".properties";
char sep = System.getProperty("file.separator").toCharArray()[0];
propFileName = System.getProperty("user.home") + sep + propName;   // always user.home
```

Config is read from `/WEB-INF/` only as a fallback when the user-home read produced nothing:

```java
// Startup.java (develop ~line 100)
if (p.isEmpty()) {
    // if the file not found in the user root, look in the WEB-INF directory
    p.readFromFile("/WEB-INF/" + propName);
}
```

But the generated key is written back to `propFileName` (user-home), regardless of source:

```java
// Startup.java (develop ~line 139)
if (Objects.isNull(secretKey)) {          // PR #2887 widens this to: || secretKey.isBlank()
    secretKey = EncryptionUtils.generateSecretKey();
    p.saveProperty(propFileName, EncryptionUtils.SECRET_KEY_ENV_VAR, secretKey);  // → user.home
}
```

`saveProperty` opens the target in **append mode**, creating it if missing, and writes a single `key=value` line:

```java
// src/main/java/io/github/carlos_emr/CarlosProperties.java:545-549
try (FileWriter writer = new FileWriter(propertyFile, true)) {
    writer.write("\n" + key + "=" + value + "\n");
    carlosProperties.setProperty(key, value);
}
```

## Reproduction (the only deployment shape that triggers it)

Requires **both**: configuration deployed in `/WEB-INF/` (no user-home properties file) **and** no encryption key set in that config.

- **Boot 1:** user-home read fails → `/WEB-INF/` config loads (DB creds, etc.) → no key present → a new key is generated → a **stub** user-home file `~/<context>.properties` is created containing only the encryption key line. The in-memory singleton is updated, so this boot works correctly.
[...]

PR #3179, opened 2026-07-13
fix: stop key-only user-home stub from shadowing WEB-INF config (#2969)

adulegendary:
## Description
Fix startup configuration loading when a user-home properties file contains only a generated encryption key.

## Problem
A previous startup could load configuration from /WEB-INF/ without an encryption key, generate one, and persist it to the user-home properties file. On the next startup, this key-only stub was detected as a non-empty configuration, causing the application to skip loading /WEB-INF/ settings.

As a result, the application started with an encryption key but missing required database configuration.

## Solution
- Detect when the user-home properties file contains only the generated key.
- Reload and merge /WEB-INF/ configuration in this case.
- Preserve the generated encryption key while restoring the full application configuration.

## Testing
- Verified startup behavior with a key-only user-home properties file.
- Confirmed /WEB-INF/ configuration is loaded correctly.

## Summary by Sourcery

Ensure startup loads WEB-INF configuration when the user-home properties file contains only a generated encryption key, while preserving any existing real key.

Bug Fixes:
- Treat a user-home properties file containing only the generated encryption key as incomplete configuration so WEB-INF settings are still loaded and merged.
- Preserve an existing user-home encryption key when merging fallback WEB-INF configuration that includes its own placeholder key.

Tests:
- Add unit and regression tests covering key-only user-home stubs, two-boot startup behavior, and protection against WEB-INF placeholder keys overwriting real user-home keys.

---
## Summary by cubic
Stop a key-only user-home properties file from suppressing `/WEB-INF` DB config on startup. Previously, a generated key or packaged defaults with `db_username` made startup skip `/WEB-INF` and boot without DB settings; now we base the fallback on deployment-supplied DB config and preserve the real key.

- Decide the `/WEB-INF` fallback from deployment-supplied config only: read `carlos_override_properties` and the user-home file via `Startup.readDeploymentProperties`, not the preloaded `CarlosProperties` singleton.
- When merging `/WEB-INF/<context>.properties`, keep the existing user-home `encryption.util.secret.key`; a `/WEB-INF` placeholder cannot overwrite it.
- Add `Startup.hasDatabaseConfiguration(Properties)`, `readDeploymentProperties(...)`, and `loadIfPresent(...)`; expand tests to cover the two-boot flow, packaged-defaults case, key-overwrite guard, and missing-config abort using new `/WEB-INF/carlosmerge*.properties` fixtures.

<sup>Written for commit 4c81ae25b0b44f03129b50c2219898f9652f956c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3179?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


adulegendary:
Hi @Ben-Heerema , @yingbull — following up to see if there's any update on merging this. Anything I can do on my end to help it along? Thanks!

Ben-Heerema:
@adulegendary Sorry, we have been busy trying to get the next MVP ready. I will look into it now, thanks for your patience!

Ben-Heerema (dismissed):

Ben-Heerema:
@adulegendary  Thanks for the PR. The main fix looks directionally right, but I think there is one edge
 case that should be handled before merge. My AI tried to do it without being asked, so I apologize for stepping on toes. I reverted it to let you do it. 

The edge case:
When the user-home file contains only `encryption.util.secret.key`, `Startup` now falls back to loading `/WEB-INF/<context>.properties` into the same `CarlosProperties` instance. If the WEB-INF properties file also contains `encryption.util.secret.key` as a default/placeholder value, `Properties.load()` can overwrite the previously generated user-home key. That would risk breaking decryption for data encrypted after the first startup.

Could you update the merge path so the user-home key is preserved when loading WEB-INF fallback config, and add a regression test where WEB-INF contains DB config plus a different placeholder key? The final assertions should show that DB config is loaded from WEB-INF, but the key remains the one from the user-home stub.

The other Gemini comment about reflection/test isolation does not look actionable to me:
  `StartupUnitTest` is already annotated with `@Isolated`, and `SECRET_KEY_SPEC` is not `static final`.
  
Thanks for your interest in this project :)

adulegendary:
Thanks @Ben-Heerema for point that out. Sorry, I closed this by accident. The edge case you flagged is handled: the WEB-INF fallback now preserves the already-generated user-home key instead of letting a placeholder key in /WEB-INF/ overwrite it. Added a regression test (shouldKeepUserHomeKey_whenWebInfFallbackHasDifferentPlaceholderKey) where WEB-INF carries DB config plus a *different* placeholder key — it asserts DB config loads from WEB-INF while the key stays the user-home stub's. Verified it fails without the fix and passes with it..

Ben-Heerema:
@coderabbitai review
@cubic-dev-ai review

Ben-Heerema:
@adulegendary looks like the CI has some concerns, mind looking into those? thanks :)

adulegendary:
> Caution
> 
> The consumer version of Gemini Code Assist on GitHub has been sunset. All code review activity has officially ceased.

Thanks for the review @Ben-Heerema ! Addressed both findings:

P1: the WEB-INF fallback now keys on whether a usable DB config is actually present (db_username), instead of inspecting the pre-populated CarlosProperties singleton. This also covers the case where classpath /carlos.properties carries only non-DB boilerplate,  which a purely user-home-file check would have missed.
P3: corrected the inaccurate pre-fix comment.
Added a regression test reproducing the singleton-pollution scenario; full StartupUnitTest is green (20/20).

Ben-Heerema:
@cubic-dev-ai review
@coderabbitai review

adulegendary:
The error cateched was confirmed and it’s a valid P1. The issue was that CarlosProperties already loads db_username=root from the packaged defaults, so our check always thought DB config existed and skipped the /WEB-INF/ fallback.

Fixed in c8199f0 by checking the actual deployment config (override + user-home) instead of the shared singleton. I also added a regression test that reproduces the issue and updated two existing tests that were relying on the packaged defaults to pass.

@Ben-Heerema, one behavior change to note: if no real config exists in any location, startup now fails fast instead of using the packaged defaults.

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
@adulegendary Looks like the CI has some concerns. Thanks for your work thus far, mind addressing them?

adulegendary:
Apologies for the delay. @yingbull  The PR is updated and ready for review.

All findings are addressed:

* Preserved the user-home encryption key during the WEB-INF merge.
* Fixed the database configuration check.
* Made unreadable configuration files fail with a clear error.
* Prevented reading the user-home file twice.

All 23 `StartupUnitTest` tests pass, and the branch is up to date with `develop`.

PR #3192, opened 2026-07-18
chore: remove JitPack repository by vendoring OpenPDF 3.0.3 into local_repo

yingbull:
## Description

Removes the `jitpack.io` repository from `pom.xml` entirely, without changing the runtime classpath by a single byte.

JitPack builds artifacts from source on demand and can serve different bytes (and thus different SHA512 hashes) between fetches, which caused spurious dependency-lock failures and let it shadow Maven Central artifacts under alternate groupIds. `ultrabuk-htmltopdf-java` had already been vendored into `local_repo`, but JitPack was still silently serving two artifacts: `com.github.librepdf.openpdf:openpdf:3.0.3` and `:openpdf-html:3.0.3`.

Those JitPack-style coordinates (with the extra `.openpdf` groupId segment) are deliberate and must be kept: they let modern OpenPDF 3.x (`org.openpdf.*`) coexist with legacy `com.github.librepdf:openpdf:1.3.43` (`com.lowagie.*`, required by JasperReports), since Maven allows only one version per groupId:artifactId. Repointing to Central's official coordinates would break one of the two consumers — so instead this PR vendors the JitPack artifacts byte-for-byte, exactly as was done for ultrabuk.

Changes:
- Vendor `openpdf` / `openpdf-html` 3.0.3 jars + POMs and the `openpdf-parent` POM into `local_repo`, byte-identical to the SHA512 hashes already recorded in `dependencies-lock.json` (verified before commit).
- Remove the `jitpack.io` repository declaration from `pom.xml`.
- Remove the dependency-lock integrity/version exemption for `ultrabuk-htmltopdf-java` — all formerly JitPack-served artifacts now have deterministic bytes, so strict version + SHA512 checking applies to the entire dependency tree (previously tracked as long-term follow-up in the plugin config comment).
- Update stale JitPack-related comments in the `jspc` and `skip-dependency-lock` profiles (profiles themselves unchanged — CI workflows reference them by name).

`dependencies-lock.json` is unchanged: the resolved dependency tree is identical before and after.

## Related Issues

Related to the "JitPack exclusion" follow-up noted in the dependency-lock plugin configuration and item 2 of the follow-up list in `docs/pdf-dependency-consolidation-plan.md` (removes the JitPack repo; replacing ultrabuk itself remains future work).

## How Was This Tested?

- With a completely empty `~/.m2`, `mvn se.vandmo:dependency-lock-maven-plugin:1.1.1:check` resolved the full tree from Maven Central + `local_repo` only — zero requests to jitpack.io — and passed: "Actual dependencies matches locked dependencies".
- Full `mvn -B -T 1C -DskipTests=true -Dcheckstyle.skip=true package` (the same command the carlos-podman Containerfile runs, but **without** `-Pskip-dependency-lock`) succeeded and produced the WAR; the strict lock check ran during the build and passed.
- Verified the vendored jars' SHA512 hashes match the existing `dependencies-lock.json` entries exactly, and that the vendored POMs declare no repositories of their own.

## Checklist

- [x] My commits are signed off for the [DCO](https://developercertificate.org/) (`git commit -s`) — via DCO remediation commit 1c6cbf42 adding Signed-off-by for 2a43dc22 (history rewriting is blocked in the sandboxed environment)

https://claude.ai/code/session_0161zmp3S9D3RgGdYw3FUJ3P


yingbull:
@copilot+claude-opus-4.8 what should be done re socket security warning for cve

yingbull:
@copilot+gpt-5.5 Check if this issue is valid — if so, understand the root cause and fix it. At local_repo/com/github/librepdf/openpdf/openpdf-html/3.0.3/openpdf-html-3.0.3.pom, line 18:

<comment>The published POM gives consumers conflicting license metadata: it names LGPL 2.1-or-later but links directly to the LGPL 3.0 text. Using the canonical LGPL 2.1 URL here (or explicitly naming LGPL 3.0) keeps dependency reports and compliance tooling aligned with the stated license.</comment>

<file context>
@@ -0,0 +1,141 @@
+  <licenses>
+    <license>
+      <name>GNU Lesser General Public License (LGPL), version 2.1 or later</name>
+      <url>https://www.gnu.org/licenses/lgpl-3.0.html</url>
+    </license>
+  </licenses>
</file context>

yingbull:
Rebase commit -s for dco

yingbull:
@copilot+gpt-5.5 Rebase commit -s for dco

yingbull:
Confirming DCO sign off for all commits

PR #3427, opened 2026-08-12
fix(casemgmt): refresh issues after CPP saves

### LINKED ISSUE #3422: [Bug]: Saving CPP encounter items throws a JavaScript error and skips issue refresh
## Description

Saving an eChart CPP item persists the note, but the post-save callback throws a JavaScript exception and does not complete the Unresolved Issues refresh.

This affects Social History, Medical History, Family History, Ongoing Concerns, Risk Factors, Reminders, and Other Meds because they share the same save callback.

## Steps to Reproduce

1. Log in with a local synthetic development account.
2. Open a patient encounter.
3. Select **Add Item** in Social History.
4. Enter text and save the item.
5. Observe the browser console and the Unresolved Issues panel.

## Expected Result

The CPP item saves without a browser error and the Unresolved Issues panel refreshes.

## Actual Result

The item is persisted and appears in Social History, but the callback throws:

```text
CarlosAjax onSuccess error: TypeError: Cannot read properties of null (reading 'value')
    at onIssueUpdate (.../js/newCaseManagementView.js.jsp)
```

The failing statement is in `onIssueUpdate()`:

```javascript
var demographicNo = $("demographicNo").value;
```

The current encounter DOM does not provide that expected element ID. The HTTP save requests themselves return 200.

## Module/Area

eChart / encounter CPP items

## Technical Notes

- Source: `src/main/webapp/js/newCaseManagementView.js.jsp`, around `onIssueUpdate()`.
- Reproduced on deployed `develop` commit `b79ba7c6c6c5be5ce8f6db1988987cc66f17f324`.
- The relevant source remains unchanged in current `origin/develop` at the time of filing.
- The existing eChart Playwright check only opens the Social History editor; it does not save an item, so it does not catch this failure.

## Acceptance Criteria

- All seven CPP item types save without a browser console exception.
- Unresolved Issues refreshes after the save.
- A browser regression test saves a CPP item and verifies both persistence and the absence of console/HTTP errors.

## Checklist

- [x] Searched for similar issues
- [x] Reproduced consistently with synthetic development data
- [x] Verified that no patient data or PHI is included

Ben-Heerema:
## Summary

- use the encounter-level demographic value when the shared CPP callback refreshes Unresolved Issues, with URI-component encoding
- isolate the repeatedly evaluated note issue script and JavaScript-block encode its dynamic clinical text
- extend the eChart Playwright check to save a Social History item, verify persistence and issue refresh, reject all unexpected console/HTTP/dialog failures, and archive its own fixture
- add focused regression guards for callback behavior, fragment isolation, and output encoding

## Validation

- `mvn -Dtest=CaseManagementCppSaveRegressionTest test` (3 tests)
- `mvn -Dtest='CaseManagementAutoSaveRegressionTest,EChartLayoutRegressionTest,MeasurementsJspEncodingRegressionTest' test` (14 tests — the neighbouring suites that also pin these files)
- `npm run test:echart-playwright` (passed repeatedly against the locally deployed patch, including archival cleanup)
- `mvn -Pjspc -DskipTests package` (full JSP compilation, 0 errors)
- `node --check scripts/echart-playwright-checks.js`
- `git diff --check`
- the extended Playwright check reproduced the original `demographicNo` exception against unpatched `develop`

## Review round 2

A second review pass found no MUST-level defects. The risky-looking parts were checked and hold up:

- `demographicNo` is declared at the top of `newCaseManagementView.js.jsp` and assigned by **both** loaders (`newEncounterLayout.jsp:394`, `ChartNotes.jsp:184`), so dropping `$("demographicNo")` cannot raise a `ReferenceError`.
- The new IIFE is safe: `updatedNoteId` and `origObservationDate` are implicit-global *assignments*, not declarations, and no `'use strict'` was added, so they still reach the outer scope.
- `carlos:forJavaScriptBlock` exists as both a tag context and an EL function (`carlos-tag.tld:124`, `SafeEncode:277`).

**SHOULD — fixed**

- **`noteIssueList.jsp:462` is a behaviour change, not just an encoding change.** `noteIndex` exists twice: a scriptlet local (`String noteIndex = ""`, never reassigned) and a page-scoped EL var set by `<c:set>` on every branch. The old `"bgColour" + "<%=noteIndex%>"` therefore always produced the id `"bgColour"`, which never resolves — so note colour styling silently no-opped. Switching to `${noteIndex}` repairs that, matching what line 479 already did. Now pinned by a test so the scriptlet/EL mismatch cannot return.
- The eChart check asserted `badResponses` only *after* rethrowing the save failure, so an HTTP 500 during the save surfaced as an opaque 15 s `waitForResponse` timeout instead of the status that caused it. Reordered.
- The regression test located the end of `onIssueUpdate()` by searching for a closing brace at one specific indentation; it is now bounded by the next `function` declaration.

**COULD — fixed**

- Collapsed the duplicated `message.type() === 'error'` branch in `isBlockingConsoleMessage`.
- Corrected the `demographicNo` comment, which named only one of the two loaders that assign it.

**Withdrawn during review**

- An earlier concern that `ECHART_DEMOGRAPHIC_NO` could drift from `ECHART_SEARCH_TERM` and cause a misleading timeout does not hold: `openPatientEchart` selects the patient with `a[onclick*='DemographicEdit?demographic_no=${demographicNo}']`, so a mismatch fails earlier at that click with a clear locator error.

**Still open**

- `resolveProjectPath()` here and `projectRoot()` in #3433's test are two implementations of the same thing; worth extracting one shared test helper.
- The fragment-isolation assertion matches any `<script>` block in the file rather than specifically the AJAX-re-evaluated one.

Fixes #3422

## Summary by Sourcery

Fix CPP save issue refreshes and harden the related eChart and JSP regression coverage.

Bug Fixes:
- Refresh Unresolved Issues after CPP saves using the encounter demographic value, including proper URI encoding.
- Prevent repeated issue-fragment evaluations from conflicting and safely encode dynamic clinical text for JavaScript contexts.
- Restore note colour styling by using the page-scoped note index.

Enhancements:
- Expand the eChart browser check to save, verify, and archive a Social History item while validating the resulting issue refresh and reporting failures clearly.

Tests:
- Add regression coverage for CPP callback behavior, repeated fragment isolation, JavaScript output encoding, and note styling.


Ben-Heerema:
## Fresh manual review (MUST / SHOULD / COULD)

### MUST

- None. The callback now uses the encounter-scoped, JavaScript-encoded demographic identifier and URI-encodes it before the refresh request. The repeatedly evaluated fragment is isolated, dynamic values are contextually encoded, and the browser check cleans up its synthetic record.

### SHOULD

- **Addressed:** make the source-level regression test independent of the JVM working directory by resolving paths from the Maven project root with a bounded parent fallback.
- **Addressed:** rename both focused test methods to follow the repository BDD naming convention.

### COULD

- No additional change is worth adding to this PR. Exercising all seven CPP categories separately would add substantial browser-test time without testing distinct callback logic; Social History is an appropriate representative because all seven use the same save callback.

Re-review after the SHOULD fixes found no remaining MUST items and no further worthwhile in-scope changes.

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
Addressed both Cubic P2 findings in 43c6dc9172:

- exempt only the known `imageRenderingServlet` fixture 404 console location while preserving response-status validation
- treat unexpected console failures and dialogs from every page in the workflow as fatal, with only the explicit same-user note-lock dialog allowed

Re-review: no new MUST findings; `node --check` and `git diff --check` pass.

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
CI note: the general `tests` job is failing in the repository-wide `SurefireIncludeCoverageUnitTest`, which reports the pre-existing `DemographicExportAction42ActionRequestMethodTest` as unmatched by the Surefire selectors. That test class is on `develop` and neither it nor the coverage guard is changed by this PR. The same failure is present on the other independent PR in this series; this PR’s focused tests and checks pass.

Ben-Heerema:
Final fresh manual review of the latest code:\n\n- MUST: none\n- SHOULD: none\n- COULD: none\n\nThe callback uses the encounter demographic consistently, all dynamic callback data is encoded for its JavaScript/URI context, and browser failures are enforced without hiding unrelated errors. Focused tests pass 2/2; JSP compilation and browser regression checks pass. Cubic's final review found no issues.

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
@cubic-dev-ai review

Ben-Heerema:
@yingbull ready for review

CecilFoubert (approved):

yingbull (approved):

PR #3482, opened 2026-08-19
fix(security): authorize patient access before combining eDoc PDFs

Ben-Heerema:
## The problem

`CombinePDF2Action` accepted a list of `docNo` values and combined them with no check that the requesting provider may see the patients those documents belong to.

The `_edoc` write privilege was enforced, but that is a module-level right, not a per-patient one. Any user who could reach the combine route could name arbitrary document numbers and receive their contents merged into a single PDF.

## Changes

- Resolve each document to its demographic links through `CtlDocumentDao` and require `isAllowedAccessToPatientRecord` for each.
- Refuse with 403 when a document has no demographic link, or the provider is not permitted.
- Reject invalid document numbers with 400 before any lookup runs.
- Preserve provider-scoped and legacy link shapes deliberately: a document may be linked by provider rather than demographic, and legacy rows carry link data in a different module column. Both are matched rather than treated as unlinked.

## Why this is a separate PR

This work was written inside the outbound email archive access branch (#3204) but does not belong to it. Splitting it out lets it be reviewed and merged on its own, ahead of the archive stack.

One piece was deliberately left behind: #3204 also makes this action refuse documents that back an outbound email archive artifact. That depends on `OutboundEmailArchiveDao` and stays with the archive lane. It is the only part of the original change not carried over.

## Note on the tests

`CombinePDF2ActionTest` is renamed to `*UnitTest`. The name it had is not selected by the pom's Surefire `<includes>`, so as written these authorization tests would never have run in CI — which for a security check is worse than having none, since the file reads like coverage.

## Verification

23/23 CombinePDF2Action unit tests and 5/5 CtlDocumentDao integration tests pass. Checkstyle and diff hygiene pass. The Surefire dark-test guard still passes.

## Context

Part of the #3204 split described in the [status update on #3096](https://github.com/carlos-emr/carlos/pull/3096#issuecomment-5344912812).

---

## Summary by Sourcery

Enforce per-document access controls before combining and returning eDoc PDFs.

Bug Fixes:
- Authorize every eDoc in a combined PDF request against patient, provider, facility, and program visibility before serving its contents.
- Reject malformed, missing, unauthorized, and oversized document requests with appropriate HTTP errors.

Enhancements:
- Support bulk document-link lookup while preserving demographic, provider-scoped, and legacy document associations.
- Prevent combined PDFs from being cached and record an audit entry for each document read.

Tests:
- Add DAO integration coverage and unit tests for document validation, authorization scopes, response security, and audit logging.

---
## Summary by cubic
Authorizes each eDoc before combining PDFs so callers can no longer read documents outside their patient access. The combine route previously merged any `docNo` list with only `_edoc` write; it now validates each document, checks demographic and provider scopes, and enforces facility/program restrictions, returning 400/403/404/413 where appropriate.

- Adds `CtlDocumentDao.findByDocumentNos(Collection<Integer>)` for bulk link lookup.
- Rejects invalid, duplicate, or excessive (over 100) document numbers early.
- Requires `isAllowedAccessToPatientRecord` for demographic links; provider-linked documents pass only when public or owned by the caller.
- Enforces `FILTER_ON_FACILITY` and program-restricted document visibility.
- Hardens the PDF response with no-store caching and audits each document read.
- Adds unit and DAO integration tests; the original test class was renamed so CI runs it.
- No DB or config changes; refusing documents that back outbound email archive artifacts remains in #3204.

<sup>Written for commit 0d45af73583eaa2bc9c5a5d006d10bf9be59635b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/main/java/io/github/carlos_emr/carlos/commn/dao/CtlDocumentDao.java
+++ b/src/main/java/io/github/carlos_emr/carlos/commn/dao/CtlDocumentDao.java
@@ -31,6 +31,7 @@
  */
 package io.github.carlos_emr.carlos.commn.dao;
 
+import java.util.Collection;
 import java.util.List;
 
 import io.github.carlos_emr.carlos.commn.model.CtlDocument;
@@ -40,5 +41,16 @@ public interface CtlDocumentDao extends AbstractDao<CtlDocument> {
     public CtlDocument getCtrlDocument(Integer docId);
 
     public List<CtlDocument> findByDocumentNoAndModule(Integer ctlDocNo, String module);
+
+    /**
+     * Finds every module link for the supplied document numbers.
+     * Null document numbers and duplicate values are ignored. A null, empty, or
+     * all-null collection returns an empty list. The returned list is never null
+     * and has no defined ordering.
+     *
+     * @param documentNos document numbers whose module links should be returned
+     * @return all matching module links, or an empty list when none match
+     */
+    public List<CtlDocument> findByDocumentNos(Collection<Integer> documentNos);
 
 }
--- a/src/main/java/io/github/carlos_emr/carlos/commn/dao/CtlDocumentDaoImpl.java
+++ b/src/main/java/io/github/carlos_emr/carlos/commn/dao/CtlDocumentDaoImpl.java
@@ -31,9 +31,11 @@
  */
 package io.github.carlos_emr.carlos.commn.dao;
 
+import java.util.Collection;
 import java.util.List;
 
 import jakarta.persistence.Query;
+import jakarta.persistence.TypedQuery;
 
 import io.github.carlos_emr.carlos.commn.model.CtlDocument;
 import org.springframework.stereotype.Repository;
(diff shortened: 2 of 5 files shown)


Ben-Heerema:
@coderabbitai review

Ben-Heerema:
CodeRabbit follow-up complete:

- All three actionable inline findings were addressed in `0d45af7358` and their threads are resolved.
- The latest CodeRabbit pass reports no actionable merge-blocking risk and marks the changed code LGTM.
- The remaining 80% docstring-coverage warning is not being applied: it measures all 50 touched functions, including private helpers and test methods, and would add broad documentation churn outside this security fix. The new public contracts in scope are documented (`CombinePDF2Action.execute()` and `CtlDocumentDao.findByDocumentNos(...)`).
- The full CI, test, Sonar, and security-check suite is green.

No unresolved review threads remain.

yingbull (approved):

PR #3481, opened 2026-08-19
test(email): cover attachment settings request validation

Ben-Heerema:
The attachment-settings request factory now has regression coverage proving that sender validation, subject sanitization, content limits, and safe encryption/manual-send defaults are applied to HTTP input before it becomes email settings. Remove one stale Surefire baseline entry whose class has already been renamed and enabled.

### Reconciliation with develop

Merged `origin/develop` at `8d77370162` into the existing `fix/email-attachment-settings-dark-test` branch and resolved both conflicts. The original email-test rename and corrected 254/255-character fixtures already landed in develop's broader test-enablement work. Preserve those fixtures and the current removal of manual-password sanitization; do not reintroduce obsolete tests or the old unmatched-test backlog.

The final diff against develop contains only:
- Five request-factory test cases in `EmailAttachmentSettingsUnitTest`.
- Removal of the obsolete `DemographicExportAction42ActionRequestMethodTest` baseline entry; its replacement `*UnitTest` remains selected and tested.

### Validation

- Focused Maven run: **38 tests, zero failures/errors/skips** (35 email-settings tests, two demographic export-method tests, and the Surefire discovery guard). Dependency-lock verification and repository Checkstyle pass; BDD naming passes.
- Mutation checks fail when sender validation is bypassed, subject sanitization is bypassed, automatic sending defaults become unsafe, or the email test is renamed back to the old unselected name. Passing baselines are restored after each probe.
- Installed Ubuntu 26.04 VM: `scripts/email-recovery-playwright-checks.js` passes with no email sent, and `scripts/application-health-playwright-checks.js` passes. Application/runtime, database, packaging and Maven sources are byte-identical to the previously installed and validated `9527846368` package, so this test-only update requires no replacement application binary.
- Deployment health initially flagged only the disposable VM's backup age exceeding two days while it had been stopped. The configured backup job completed and verified its repository successfully; the repeated `carlos-ctl check` passes all checks. No application or browser failure was found. Owned email fixtures were removed, no failed migrations remain, and application restart count is zero.
- ON/common: 28 active Flyway versions; BC/common: 25. No normalized version collisions. All production/schema/configuration/script assets match develop.
- VM uses 6 GiB RAM and a 2 GiB application heap, with host/guest memory guards. Host compilation finishes before VM startup.

### Findings and review status

[...]

PR #3489, opened 2026-08-19
test: un-darken 118 test classes Surefire never selected

Ben-Heerema:
## What was hidden

The pom's Surefire `<includes>` only select `*UnitTest`, `*IntegrationTest`, `*RegressionTest` and `*Tests`, plus a few package paths. A plain `*Test` class anywhere else compiles, passes under an explicit `-Dtest=` run, and **is never executed by `mvn test`**.

`surefire-unmatched-baseline.txt` recorded 132 such classes as known backlog. `SurefireIncludeCoverageUnitTest` only fails on *growth*, so the backlog itself stayed invisible.

It was hiding **1,021 test methods**.

## Measured before touching anything

I ran all 131 dark classes explicitly first, because renaming blind would just turn an invisible problem into a red build:

| | |
|---|---:|
| tests executed | 1,078 |
| passed | 1,045 |
| failed | 33 |
| classes holding those 33 failures | **8** |

So this is not 131 broken suites. It is ~123 working suites nobody was running, and a small number of genuinely broken ones.

## What this PR does

Renames the 118 that pass and are not already in flight elsewhere, and drops their baseline entries. **865 tests now run in CI that did not before.**

No production code changes. No test bodies changed beyond the class name.

## What it deliberately leaves behind

**The eight classes holding all 33 failures** — `FindNextAvailableSlot2ActionTest` (24 of them), `FilterRegistryTest`, `SystemMessage2ActionTest`, `StrutsAdminConfigTest`, `RxShowAllergy2ActionTest`, `TargetClinicalJspI18nTest`, `OscarResourcesBundleParseTest`, `EmailAttachmentSettingsTest`. Each needs a real fix, not a rename; enabling them as-is would just make CI red. `EmailAttachmentSettingsTest` is already fixed in #3481 — its assertion built a 253-character string and asserted it was 254, so it could never have passed.

**Two in flight elsewhere** — `CombinePDF2ActionTest` and `ManageDocument2ActionTest`, renamed or modified on the branches for #3482 and #3484. Renaming them here would collide.

**Three whose `*UnitTest` name is already taken** — `DocumentAttachmentManagerImplTest`, `EctMeasurements2ActionTest`, `ImageRenderingServletTest`. A `*Test` and a `*UnitTest` of the same name is a duplicate-suite question, not a rename.

**`QuickTest`** — it duplicates `PreventionItemsXmlUnitTest`, whose own Javadoc says it was migrated from it. It should probably be deleted rather than enabled under a name that means nothing. That is a maintainer's call, not a sweep's.

## Why this matters

Six separate defects surfaced from this backlog in a single day of email work, including two that a `*Test` name had kept invisible: a security test file for `CombinePDF2Action` that read like coverage but never ran, and an assertion that was arithmetically impossible. A test that is wrong and silently absent is indistinguishable from one that is right.

## Verification

All 118 renamed classes run green — 865 tests, 0 failures, 3 skipped. The Surefire coverage guard still passes.

Expect CI on this PR to take noticeably longer than usual: that is the 865 newly-running tests, and is the point.

---

## Summary by Sourcery

Enable the previously hidden passing test suites in Surefire and strengthen verification that HL7 metadata logging does not expose patient information.

Enhancements:
- Rename 118 previously unselected test classes to Surefire-compatible unit or integration test names so they run in the standard test suite.
- Improve the HL7 logging test to verify captured metadata logs exclude patient-identifying information while avoiding downstream file and import operations.

Documentation:
- Update path-validation and UI test documentation to reference the renamed test classes.

Tests:
- Expand CI coverage by enabling 118 previously unmatched test classes, adding 865 passing tests while retaining the known failing and conflicting suites for follow-up.

Chores:
- Remove the enabled test classes from the Surefire unmatched baseline and update related source references and Javadocs.

---
## Summary by cubic
Renames 118 previously unselected test classes to Surefire-compatible `*UnitTest` or `*IntegrationTest` names, so `mvn test` now runs 906 additional tests instead of silently skipping them. No production code changes; CI takes noticeably longer.

- Rewrote the HL7 MSH-logging test in `SubmitLabByForm2ActionUnitTest` to stop after the metadata log statement and assert from captured logs that no PHI is emitted.
- Updates related Javadocs, test references, and path-validation documentation for the new names.
- Removes the corresponding entries from `surefire-unmatched-baseline.txt`; the Surefire coverage guard still passes.
- Leaves eight failing suites, two suites changed on other branches (including `EmailSend2ActionUnitTest`, which took develop's independently expanded version), three name collisions, and duplicate `QuickTest` for follow-up.
- Verified all renamed suites: 906 tests passed, with 3 skipped.

<sup>Written for commit e21aef0116d0a98da2144c62625fe505b79e1d72. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3489?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Ben-Heerema:
@coderabbitai review

Ben-Heerema:
@coderabbitai review

yingbull (dismissed):

PR #3505, opened 2026-08-20
fix: repair the seven remaining dark test classes and enable them

Ben-Heerema:
Stacked on #3489, which renamed the 118 dark test classes that already passed and deliberately left these seven behind — they needed real fixes, not renames.

All seven now pass and come off the baseline. That leaves **7 entries in `surefire-unmatched-baseline.txt`**, none of them failures: they are duplicate-suite and delete-candidate questions listed in #3489.

## Two were real production defects

The dark tests had been catching these all along, with nobody watching.

**Mojibake in four locale bundles.** `oscarResources_en` carried a raw UTF-8 em-dash, and the `es`/`fr`/`pl`/`pt_BR` bundles carried **667 raw non-ASCII bytes** between them. `Properties.load(InputStream)` decodes ISO-8859-1, so every accented character in those files was rendering as mojibake in that locale's UI — the `en` case being an Ontario billing error message shown when code totals fail to parse. All now `\uXXXX`-escaped, which is what the bundle format requires. I verified the escapes decode back to the intended text.

**A dead script include.** `securitydelete.jsp` loaded `/js/global.js` while containing no JavaScript at all — no `onclick`, no `function`, nothing. It is a result-only display page, so the include was pure weight.

## Five were test-side

| Class | Problem |
|---|---|
| `FilterRegistryTest` | Did not extend `CarlosUnitTestBase`, so constructing `ExistingAppointmentFilter` — which resolves `ScheduleManager` in a field initialiser — hit a null bean factory |
| `SystemMessage2ActionTest` | Invoked action methods directly, outside a Struts dispatch, so `ActionSupport.getText()` had no `Container` and NPE'd |
| `RxShowAllergy2ActionTest` | Asserted the `_allergy` read check happens once; it happens twice |
| `TargetClinicalJspI18nTest` | Required every target JSP to declare the bundle, including ones that render nothing |
| `FindNextAvailableSlot2ActionTest` | Described an implementation that no longer exists (see below) |

Two of those deserve a note.

**`RxShowAllergy2ActionTest`** — the read check really does fire twice: `execute()` gates the dispatcher and `reorder()` gates the handler it dispatches to. That duplication is deliberate defence in depth, since `reorder()` is reachable directly. So the assertion pins `times(2)` rather than being relaxed to `atLeastOnce()`, which would have hidden a future change in either gate.

**`TargetClinicalJspI18nTest`** — the JSP it flagged, `demographiceditdemographic.jsp`, is a twelve-line `<jsp:forward>` compatibility stub. Requiring a bundle declaration there would be dead code. The assertion now applies only to pages that actually use `fmt:message`, which is what it was protecting in the first place.

## `FindNextAvailableSlot2ActionTest` was rewritten, not repaired

It held **24 of the 33 failures**. It reflected on private helpers like `timeStrToMins` and drove a day-by-day appointment scan the action used to perform inline. That work now lives in `NextAppointmentSearchHelper`, so the helpers it reached for simply do not exist — every failure was `NoSuchMethodException`.

The replacement covers what this action still decides for itself: authorization, parameter handling, provider fan-out, global sorting across providers, ordinal selection with its short-result fallback, and response shape. It is deliberately smaller than what it replaces, because it does not restate behaviour that now belongs to the search helper and is that class's to test.

Worth recording: this drift was only findable because the class now runs. It stayed invisible for exactly as long as the `*Test` name kept Surefire from selecting it.

## Verification

51 tests across the seven classes, all green, plus the Surefire coverage guard. No live deployment — local test runs only.

Depends on #3489.

---

## Summary by Sourcery

Enable and repair the seven remaining dark test classes while fixing the production issues they exposed in locale resource encoding and an unnecessary script include.

Bug Fixes:
- Correct locale resource encoding so accented and other non-ASCII messages render correctly.
- Remove the unused global JavaScript include from the security deletion page.

Enhancements:
- Bring the seven previously skipped test classes into the active unit-test suite and update their coverage for current application behavior.

Tests:
- Replace obsolete appointment slot action tests with coverage for authorization, parameter validation, provider aggregation, result selection, fallback behavior, and JSON responses.
- Correct test setup and expectations for filter dependencies, Struts text lookup, allergy authorization checks, and JSP bundle requirements.

Chores:
- Remove the repaired test classes from the unmatched-test baseline.

---
## Summary by cubic
Enables the seven remaining dark test classes that Surefire previously skipped by renaming them to selected patterns and updating them for current application behavior. It also escapes raw non-ASCII characters in the `es`, `fr`, `pl`, and `pt_BR` locale bundles so accented text no longer renders as mojibake, and removes the unused `/js/global.js` include from `securitydelete.jsp`.

**Test coverage**

- Replaces obsolete appointment action tests with coverage for authorization, parameter handling, provider fan-out, global sorting, ordinal fallback, and JSON responses.
- Fixes test setup and assertions for filter dependencies, Struts text lookup, allergy authorization checks, and JSP bundle usage.
- Removes the seven repaired classes from `surefire-unmatched-baseline.txt`; the remaining seven entries are duplicate-suite or delete-candidate follow-ups.

<sup>Written for commit 0ea74556dd024d0ad057ea35d4b415b220d30b2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3505?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Ben-Heerema:
@coderabbitai review

yingbull (dismissed):

PR #3506, opened 2026-08-20
fix(i18n): stop the Norwegian language name rendering as mojibake

Ben-Heerema:
## The bug

`spoken_languages_codes.properties` held `Bokmål` as raw UTF-8. `SpokenLangProperties` loads the bundle through `Properties.load(InputStream)`, which decodes **ISO-8859-1**.

```
file bytes        b'Bokm\xc3\xa5l'
ISO-8859-1 decode  BokmÃ¥l          <- what the patient language picker has been showing
```

Escaped to `Bokmål`, which is what the properties format requires for anything outside ASCII.

## How it was found

While fixing the same defect class in the `oscarResources` bundles (#3505). That one is guarded by `OscarResourcesBundleParseTest` — but it only scans `oscarResources_*`, so this file was never checked by anything.

## Tests

Two, and they check different things on purpose:

- **The decoded value**, through the real accessor. It was the *decode* that was wrong, not the character, so asserting file bytes alone would miss the point.
- **The whole file is ASCII-only.** Any future accented language name added as raw UTF-8 fails the same way, and would otherwise only be noticed by someone reading the dropdown.

## Related, deliberately not fixed here

`mspEditCodes.properties:661` has a different problem:

```
ZG=FOR USE BY AN ADJUDICATOR TO BY-PASS THE 40<U+FFFD> KEY ENTRY ENTRY CHARGE
```

That is `U+FFFD`, the replacement character — the data was already corrupted upstream rather than merely mis-encoded, so the original character is lost. Context suggests `40¢`, and the duplicated "ENTRY ENTRY" looks wrong too, but both need a decision rather than a guess.

---

## Summary by Sourcery

Fix Norwegian language-name decoding in the spoken-languages resource and add safeguards against future encoding regressions.

Bug Fixes:
- Correct the Norwegian Bokmål language name so it displays properly instead of as mojibake.

Tests:
- Add regression coverage for decoded language names and enforce ASCII-only content in the spoken-languages properties bundle.

---
## Summary by cubic
Fixes mojibake for the Norwegian Bokmål name in the patient language picker by escaping the å in the spoken-languages bundle so Java’s ISO-8859-1 `Properties.load` decodes it correctly. Old behavior: “BokmÃ¥l”; new behavior: “Bokmål”; no other behavior changes.

- Replaces raw UTF-8 with `\u00e5` in `NOB=Bokm\u00e5l, Norwegian` to follow the `.properties` format.
- Adds regression tests: one asserts the decoded value via `SpokenLangProperties`; one enforces ASCII-only source in `spoken_languages_codes.properties` to prevent future mojibake.
- No migration or config changes required.

<sup>Written for commit 34cbe4454ee89890bde179bb5a8fff6c0d6decaf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3506?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Ben-Heerema:
@coderabbitai review
@cubic-dev-ai review

Ben-Heerema:
@cubic-dev-ai review
@coderabbitai review

yingbull (approved):

PR #3557, opened 2026-08-29
fix(deb): keep the EMR up when three packages restart it in one transaction

yingbull:
## The defect

`carlos-emr.service` carries `StartLimitBurst=6` over `StartLimitIntervalSec=30min`
to catch a crash loop. But **three** packages each restart that one unit inside a
single apt transaction:

| package | why it restarts `carlos-emr.service` |
|---|---|
| `carlos-emr` | its own configure step starts the app |
| `carlos-emr-drugref` | so Tomcat picks up the new DrugRef context |
| `carlos-emr-eform-renderer` | so the JVM re-reads the render properties |

The renderer's restart arrived with the package itself in #3543, taking the
per-transaction count from two to three.

Three plus three is six, so the **documented** workflow — install, run the
post-install `carlos-ctl restart`, then apply a hotfix upgrade inside the same half
hour — spends the whole budget. systemd refuses the last restart with
`start-limit-hit`, the unit ends `failed`, and the front door serves **502 with no
automatic recovery**.

`apt` still exits **0**, so nothing in the transaction reports a problem. The only
tell is the renderer's `WARNING: the EMR did not restart`. Recovery needs a
hand-run `systemctl reset-failed`, which the upgrade path never mentions.

## Reproduction (packaged install, LXD VM, Ubuntu 26.04)

```
$ apt-get install -y --reinstall ./carlos-emr_*.deb ./carlos-emr-drugref_*.deb \
                                 ./carlos-emr-eform-renderer_*.deb
carlos-emr-eform-renderer: WARNING: the EMR did not restart — eForm PDF
$ systemctl is-active carlos-emr
failed                       # Active: failed (Result: start-limit-hit)
$ curl -sk -o /dev/null -w '%{http_code}\n' https://localhost/carlos/
502
```

`journalctl -u carlos-emr` shows `Start request repeated too quickly.`

## The fix

A restart we asked for on purpose is not a crash loop, so clear the start counter
before every package-driven lifecycle action and before `carlos-ctl start|restart`.
`systemctl reset-failed` clears the failed state and the start-rate counter together.

The two dependent postinsts now also act on a unit an earlier package in the same
transaction left `failed` — their guard was `is-active` only, so once the limit was
hit they silently skipped the restart and left the EMR down.

**Crash-loop protection is unchanged.** `Restart=on-failure` with `RestartSec=15s`
still trips the same `StartLimitBurst` in about 90 seconds when the JVM will not
stay up.

## Verification

Rebuilt the three packages from this branch and re-ran the same reinstall against a
live install that already had 4 starts inside the limit window (the failing
condition):

- `carlos-emr` stays **active**; front door **200**
- **7 starts** inside the window, **zero** `start-limit-hit` refusals
- no `the EMR did not restart` warning
- `carlos-ctl check` — **all checks passed** (services, loopback-only listeners,
  TLS, live WAF 403 probe, web-service gates, DrugRef lookup, Flyway state)
- `carlos-ctl restart` works and still rejects arguments
  (`carlos-ctl restart nginx` errors as before)
- **37/37** Playwright checks pass through the WAF on **:443**

Found while validating `release/2026.08` end to end on a packaged install.

https://claude.ai/code/session_01Fnov5ppnJbdnyEDMMAJMdm

## Summary by Sourcery

Prevent intentional EMR restarts during package operations and administration from being blocked by stale systemd start-rate limits while retaining crash-loop protection.

Bug Fixes:
- Prevent package upgrades and deliberate EMR lifecycle commands from leaving carlos-emr.service down after systemd start-limit exhaustion.
- Allow dependent package post-install actions to recover the EMR when an earlier package has left the unit failed.

Enhancements:
- Preserve crash-loop protection while distinguishing intentional service starts and restarts from automatic failures.
- Ensure database credential rotation can restart the EMR reliably after updating credentials.

---
## Summary by cubic
Fixes upgrades leaving the EMR down behind a 502 when three packages restart the same unit in one apt transaction and exhaust its systemd start limit. Package-driven restarts, `carlos-ctl start|restart`, and `carlos-ctl rotate` now clear the start counter first; the dependent postinsts also recover a unit left failed by an earlier package in the same transaction. Crash-loop protection is unchanged.

<sup>Written for commit ca8d359001a2aff2f461332ac66f5212d54a1a96. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3557?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/debian/assets/carlos_ctl/cli.py
+++ b/debian/assets/carlos_ctl/cli.py
@@ -12,7 +12,6 @@ share a concept; see the package docstring in __init__.py.
 """
 
 import os
-import subprocess
 import sys
 from typing import List, Optional
 
@@ -111,18 +110,11 @@ def _cmd_lifecycle(verb: str, argv) -> int:
             f"(for other units use systemctl directly)")
     need_root(verb)
     if verb in ("start", "restart"):
-        # carlos-emr.service carries StartLimitBurst=6/StartLimitIntervalSec=30min
-        # to catch a crash loop. Config changes are applied by restarting, and a
-        # package upgrade restarts the unit three times of its own accord, so an
-        # operator iterating on carlos.properties can spend that budget without
-        # anything being wrong — systemd then refuses with "start-limit-hit" and
-        # the EMR stays down until someone runs `systemctl reset-failed` by hand.
-        # A restart asked for explicitly is never a crash loop, so clear the
-        # counter first. Genuine crash looping is still caught: Restart=on-failure
-        # with RestartSec=15s burns the same budget in about 90 seconds.
-        subprocess.run(["systemctl", "reset-failed", "carlos-emr.service"],
-                       stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
-                       check=False)
+        # An operator asking for a restart is never a crash loop; clear the
+        # start-rate counter so systemd cannot refuse it. See
+        # util.reset_emr_start_limit for why this is needed and why it does not
+        # weaken crash-loop protection.
+        util.reset_emr_start_limit()
     os.execvp("systemctl", ["systemctl", verb, "carlos-emr.service"])
     raise AssertionError("unreachable: execvp replaces the process")
 
--- a/debian/assets/carlos_ctl/cli.py
+++ b/debian/assets/carlos_ctl/cli.py
@@ -12,6 +12,7 @@ share a concept; see the package docstring in __init__.py.
 """
 
 import os
+import subprocess
 import sys
 from typing import List, Optional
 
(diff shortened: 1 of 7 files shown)


yingbull:
Good catch on the `rotate` gap — confirmed and fixed in ca8d359001.

The finding was right about the substance; the call is in `dbops.py:cmd_rotate`
rather than `cli.py`, which is exactly why repeating the two-line reset at each
site was the wrong shape. Centralized instead: `util.reset_emr_start_limit()`
now owns the reasoning, and both deliberate-restart sites call it —
`_cmd_lifecycle` for `carlos-ctl start|restart`, and `cmd_rotate` for the
rotation restart. `stop` is deliberately untouched, since stopping is not
start-rate limited.

Worth spelling out why this path mattered most: on refusal `cmd_rotate` does not
just fail quietly, it `die()`s with *"passwords rotated, but carlos-emr FAILED to
restart"* — so the operator is left holding freshly rotated credentials **and** a
down EMR.

Verified on a packaged install rather than by inspection. Burned the budget with
raw `systemctl` until the 7th restart left the unit `failed` and the front door
at 502, then from that state:

| command | result |
|---|---|
| `carlos-ctl rotate` | `rotated; carlos-emr restarted` — active, front door 200 |
| `carlos-ctl restart` | exit 0 — active, front door 200 |
| `carlos-ctl restart nginx` | still rejected, argument guard intact |
| `carlos-ctl check` | all checks passed |

On the review's concern about the reset weakening crash-loop protection: it does
not. Nothing calls this between systemd's *automatic* restarts, so
`Restart=on-failure` with `RestartSec=15s` still burns the same
`StartLimitBurst=6` in about 90 seconds when the JVM genuinely will not stay up.
The reset only ever fires on a restart a human or a package asked for.

PR #3558, opened 2026-08-29
fix(deb): WAF blocks saving a drawn provider signature (rule 1080)

yingbull:
## The defect

Following the WAF false positives this release already fixed (rules 1010–1070), I probed the other routes that share their payload *shapes*. One is a live, clinician-facing break with no exclusion.

**Provider Preferences → Signature Stamp → draw a signature → Save** posts the pad's `canvas.toDataURL()` output from `providerpreference.jsp:1851` as:

```
method=saveDrawn&signatureData=<encodeURIComponent(data:image/png;base64,...)>
```

to `POST /carlos/provider/providerSignatureStamp`. Single-encoded, so ModSecurity's `ARGS` transform hands CRS the literal `data:image/...;base64` prefix, and **941130 + 941170 score +5..+8 between them — over the anomaly threshold of 5 on their own**. The save is 403'd at the edge.

This is the exact data-URI false positive already excluded for:
- the consultation signature pad — rule 1020, `ARGS:signatureImage`
- eForm image saves — rules 1030/1040, `ARGS:openosp-image-link`

…on the one signature route they missed. The drawn-signature feature predates the WAF (added 2026-07-07), but the WAF ships in *this* release, so on the packaged deb the workflow is broken. Not reproducible in the devcontainer (no WAF).

## Reproduction (real UI, :443)

Draw a signature in Provider Preferences and click Save:

```
save request status : 403
status banner shown : Save failed. Please try again.
```

Root-caused from the modsec audit log — `941130/941170` on `ARGS:signatureData`, anomaly score 10 — not from the HTTP status, which alone can't distinguish a WAF 403 from the app's own CSRF-filter 403. (That distinction mattered: several routes I probed returned 403 from the CSRF filter with *zero* WAF rules matched, and are not defects.)

## The fix

New package exclusion **1080**, in the tight per-argument form of 1020 — only `ARGS:signatureData`, only POST, only this route.

- Every other argument (`method`, `providerNo`) stays fully inspected.
- Sibling routes deliberately untouched: `providerSignatureImage` only reads `providerNo`, and this route's `upload` method sends a multipart file, not a data URI.
- The value is **not** trusted on the strength of the exclusion. `ProviderSignatureStamp2Action` requires `w` access, rejects non-POST, strips the prefix, strictly Base64-decodes the remainder, caps it at 512 KB, and decodes it as an image with dimensions capped at 1000×400 before writing a `.png`.

## Verification

- **Fixed, real UI:** the drawn signature saves — *"Drawn signature saved successfully."*, HTTP 200.
- **No collateral:** every previously-covered signature/eForm/note route still passes; `providerNo`/`method` on the same route are still fully inspected; sibling routes unchanged.
- **Attacks still blocked:** `<script>` and `UNION SELECT` payloads on `ARGS:signatureData` still 403. The exclusion lifts only the base64 data-URI signatures, not real attacks.
- `carlos-ctl check` — all checks passed (incl. live WAF 403 probe).
- **37/37** Playwright checks pass through the WAF on :443.

### Also investigated, NOT a defect (documented for the record)

Rule 1010 names `ARGS:reloadUrl` (lowercase), while the eChart nav code sends `reloadURL` (capital U). ModSecurity arg matching is case-sensitive, so a `reloadURL` value containing a nested `&cmd=` *does* trip 932110 in a synthetic probe. **But the real eChart never sends such a value** — the shipped `echart-playwright-checks.js` asserts zero HTTP ≥400 across the whole encounter workflow and passes on this WAF-enabled install. No exclusion added, since widening the WAF for a false positive that doesn't occur would only reduce coverage.

Found while validating `release/2026.08` end to end on a packaged install.

https://claude.ai/code/session_01Fnov5ppnJbdnyEDMMAJMdm

## Summary by Sourcery

Allow drawn provider signatures to be saved by excluding their data payload from targeted WAF inspections on the specific POST endpoint.

Bug Fixes:
- Allow clinicians to save drawn provider signatures through the WAF without triggering false-positive blocking.

Enhancements:
- Add a narrowly scoped WAF exclusion for the signature data argument while retaining inspection of other parameters and routes.
- Document the exclusion’s scope, validation safeguards, and verification results.

---
## Summary by cubic
Fixes the WAF blocking the drawn provider signature save by adding exclusion rule 1080, so clinicians can save their signature through the front door instead of hitting a 403 error.

- The `data:image/...;base64` prefix on `ARGS:signatureData` was scoring over the anomaly threshold on CRS rules 941130/941170.
- Only `ARGS:signatureData` on POST to `/carlos/provider/providerSignatureStamp` is exempted; all other arguments and routes stay fully covered.
- The action still validates strictly: rejects non-POST, strips the prefix, strictly base64-decodes, caps at 512 KB, and checks image dimensions before writing.

<sup>Written for commit b07848f82336623aab1d0adb7095de387d9ca9ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3558?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/debian/assets/modsecurity/REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf
+++ b/debian/assets/modsecurity/REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf
@@ -223,3 +223,39 @@ SecRule REQUEST_URI "@rx ^/carlos/ws/rs/eform/(?:[0-9]+/)?json(?:[;?]|$)" \
         ctl:ruleRemoveTargetByTag=attack-rce;ARGS:json.formHtml,\
         ctl:ruleRemoveTargetByTag=attack-injection-php;ARGS:json.formHtml,\
         ctl:ruleRemoveTargetByTag=attack-protocol;ARGS:json.formHtml"
+
+# Provider signature stamp (Preferences > Signature Stamp > draw > Save).
+# providerpreference.jsp:1851 posts the pad's canvas.toDataURL() output as
+#   method=saveDrawn&signatureData=<encodeURIComponent(data:image/png;base64,...)>
+# Single-encoded, so ModSecurity's ARGS decode hands 941130/941170 the literal
+# "data:image/...;base64" prefix and — per the note on rule 1040 — those two
+# score +5..+8 between them, over the anomaly threshold of 5 on their own.
+# Verified live through the packaged front door: the clinician draws a
+# signature, clicks Save, and gets the page's own "Save failed. Please try
+# again." banner with an nginx 403 and no application log line; retrying can
+# never succeed. Same defect class as rules 1020/1030/1040, on the one
+# signature route they did not cover.
+#
+# This is NOT reproducible in the devcontainer: that stack has no WAF.
+#
+# SCOPE. Per-argument (the tight 1020/1030 form): only ARGS:signatureData, only
+# POST, only this route. Every other argument on the request — method,
+# providerNo — stays fully inspected, and the sibling routes are untouched:
+# provider/providerSignatureImage only ever reads providerNo, and the `upload`
+# method on THIS route sends a multipart FormData file rather than a data URI,
+# so neither needs an exclusion and neither gets one.
+#
+# The value is not trusted on the strength of this exclusion.
+# ProviderSignatureStamp2Action requires `w` access, rejects non-POST for every
+# mutating method, strips the data-URI prefix, strictly Base64-decodes the
+# remainder (rejecting anything that is not valid base64), caps it at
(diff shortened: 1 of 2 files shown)


yingbull:
Superseded by #3560 — same fix, recreated as a single DCO-signed commit off the current release/2026.08 (this branch's merge commit was unsigned and failed the DCO check; amend/rebase/force-push are unavailable here).

PR #3588, opened 2026-09-04
fix(billing): stop Ontario 3rd-Party/Bonus-Codes billing from 500-ing

yingbull:
## Description

Two distinct "CARLOS Error: 500" failures on the Ontario bill workflow, reported by an alpha tester on 2026.08.0-alpha10. Both were reproduced against a real Ontario install (MariaDB with the Flyway common+ON baseline and the demo dataset, Tomcat 11 with the alpha10 WAR) and both are verified fixed against the same install.

### 1. Bill-type dropdown → 3rd Party / Bonus Codes

`onChangePrivate()` in `billingON.jsp` re-opens the billing form through `/billing` for exactly two visible bill types — **3rd Party** (`PAT`, plus its `OCF`/`ODSP`/`CPP`/`STD` sub-options) and **Bonus Codes** (`BON`). Every other bill type stays on the page, which is why only those two failed.

The rebuilt URL dropped `billRegion`. `Billing2Action` then falls back to the deployment-wide `billregion` property — except it read `carlos.util.plugin.CarlosProperties`, a holder whose static field **nothing in the codebase ever populates** (`setProperties` has no caller). The fall-back therefore always evaluated to `null`, the router answered `"BC"`, and an Ontario install was handed `billingBC.jsp`:

```
Caused by: java.sql.SQLSyntaxErrorException: Table 'carlos.billingvisit' doesn't exist
    at ...commn.dao.BillingBCDaoImpl.findBillingVisits(BillingBCDaoImpl.java:92)
    at ...billings.ca.bc.data.BillingFormData.getVisitType(BillingFormData.java:175)
    at org.apache.jsp.WEB_002dINF.jsp.billing.CA.BC.billingBC_jsp._jspService
```

`billingvisit` is a BC-only table absent from the Ontario schema. The operator loses the bill they had just entered.

- `Billing2Action` now reads `io.github.carlos_emr.CarlosProperties`, the singleton actually backed by `carlos.properties` (where `carlos-ctl` writes `billregion` from the configured province).
- The dead `carlos.util.plugin.CarlosProperties` holder is removed so the same trap cannot be re-armed — it had no other reference in the repository.
- `billingON.jsp` puts `billRegion=ON` back on its self-navigation, and the two Ontario unbilled-report "Bill" links do the same, so the parameter is present and the property is only the safety net.

### 2. Bill review of a service code with no description

[...]

PR #3601, opened 2026-09-05
test(billing): browser check for Ontario 3rd-party/bonus-codes bill entry

yingbull:
## Description

#3588 fixed several "CARLOS Error: 500" failures on the Ontario bill workflow but shipped no browser coverage for the path a biller actually drives. This adds that coverage.

The routing defect: `billingON.jsp` rebuilt its own navigation query string without `billRegion`, so `Billing2Action` fell back to a holder nothing populated, answered "BC", and handed the request to `billingBC.jsp` — which queries the BC-only `billingvisit` table. The bill entry was lost along with the 500.

The check logs in, opens the Ontario bill form for a demo appointment, and switches the bill type through every option whose `onChangePrivate()` branch rebuilds the URL: `PAT`/`OCF`/`ODS`/`CPP`/`STD` (3rd party) and `BON` (bonus codes). Per type it asserts:

- `billRegion=ON` survives the self-navigation,
- the Ontario form is still the page served (`billingBC.jsp` has no `xml_billtype` control, which is exactly how the misroute presented),
- no error page renders.

It also asserts the Ontario unbilled "Bill" links are context-path prefixed and carry the region — the sibling fix in the same PR — skipping cleanly when the dataset exposes none.

**Why the region assertion is the load-bearing one.** The fall-back is the `billregion` property, and an install that has it set — the packaged Ontario install does — still routes correctly with the parameter missing. A status-only assertion would therefore pass on a re-broken page.

The check is **read-only**: it never submits a bill, so it seeds nothing and has nothing to clean up.

## Related Issues

Related to #3588 (adds the browser regression coverage for that fix). No issue is closed by this PR.

## Target Branch

Targets `release/2026.08` because it guards a fix that lives on that line and is intended to ship with the same release. It is test-only — a new `scripts/` check, its `package.json` entry, and runbook notes — so it carries no application change and leaves the branch's version and SCM metadata untouched (`2026.08.0-alpha11-SNAPSHOT` / `HEAD`).

## How Was This Tested?

Verified **both directions** against a packaged Ubuntu 26.04 `.deb` install (published `2026.08.0~alpha10` upgraded to a locally built `2026.08.0~alpha11`, nginx + ModSecurity front door, MariaDB, Ontario demo dataset):

```bash
npm run test:billing-on-third-party-playwright
```

| State of the deployed `billingON.jsp` | Result |
| --- | --- |
| As shipped (fix present) | **20/20 checks passed** |
| `billRegion=ON` token reverted out of `__commonQs` | **14/20** — exactly the six region assertions fail |

While reverted, the status and Ontario-form assertions still passed, because this install has `billregion` set and the property fall-back masked the misroute. That is precisely the masking the region assertion exists to defeat, and it is why the check does not rely on the status code alone. The JSP was restored and the suite re-run to 20/20 afterwards.

The two unbilled-"Bill"-link assertions report `SKIP` on this dataset, which exposes no unbilled items — they skip rather than record a false pass.

Not covered here: bill review (the `ServiceCodeLoader` null-description NPE). The demo dataset carries the 11 null-description Ontario `billingservice` rows but no `billing` rows, so reviewing one would mean seeding a bill and turning this into a write-with-cleanup check. That path keeps its `ServiceCodeLoaderUnitTest` coverage.

## Checklist

- [x] I preserved the target branch version/SCM metadata, or this is an explicit release-preparation PR

https://claude.ai/code/session_01GhEotcrscmDa2ztS3kzKs3

---
_Generated by [Claude Code](https://claude.ai/code/session_01GhEotcrscmDa2ztS3kzKs3)_

## Summary by Sourcery

Add browser regression coverage for Ontario bill-entry navigation across third-party and bonus-code bill types.

New Features:
- Add a read-only Playwright browser check covering Ontario third-party and bonus-code bill-type navigation.

Enhancements:
- Verify Ontario bill navigation preserves the region and remains on the correct form without rendering errors.
- Validate Ontario unbilled Bill links include the deployment context path and Ontario region, while skipping cleanly when no links exist.

Documentation:
- Document the billing browser-check configuration and required demo appointment environment variables in the Debian install validation runbook.

Tests:
- Add a package script for running the Ontario billing Playwright regression check.

---
## Summary by cubic
Adds browser regression coverage for the Ontario 3rd-party/bonus-codes bill entry fix from #3588, guarding the routing defect that dropped `billRegion` and misrouted Ontario requests to the BC form.

- Asserts `billRegion=ON` survives each navigation, the switch actually navigated with the selected type, the Ontario form is still served, and no error page renders.
- The region assertion is load-bearing: the fallback is the `billregion` property, so a status-only assertion would pass on a re-broken page.
- Form identity uses the Ontario bill-type codes (`ODP`/`BON`) because `billingBC.jsp` carries its own `xml_billtype` control.
- Requires every expected bill-type code to be present, so a disappearing one fails instead of silently dropping coverage.
- Asserts Ontario unbilled "Bill" links are context-path prefixed and carry the region; erroring report routes fail outright rather than masquerading as a dataset skip.
- Read-only: never submits a bill, so it seeds nothing and has nothing to clean up. TLS verification is only relaxed for loopback hosts.
- The PIN fill is conditional on the field being present, so the check runs even with legacy PIN disabled.
- Failure diagnostics redact PHI-correlating identifiers, keeping only routing parameters.
- Verified against a packaged Ubuntu .deb install: 30/30 pass with the fix; reverting `billRegion=ON` from the deployed JSP drops to 24/30 with exactly the six region assertions failing.
- Bill review (the `ServiceCodeLoader` null-description NPE) is not covered because the demo dataset has no `billing` rows; that path keeps its `ServiceCodeLoaderUnitTest` coverage.

<sup>Written for commit 60281b114a1dfd8b00015f1ca52ff62baba1f5a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3601?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/ui-tests/deb-install-validation.md
+++ b/docs/ui-tests/deb-install-validation.md
@@ -261,6 +261,14 @@ export PRESCRIPTION_SCRIPT_ID=45 PRESCRIPTION_DEMOGRAPHIC_NO=1
 export CONSULT_DEMO_NO=1 CONSULT_SERVICE_ID=1 CONSULT_REQUEST_ID=1
 export CONSULT_STAMP_PROVIDER_NO=999998 CONSULT_UNSIGNED_REQUEST_ID=3
 export PATIENT_LIST_FIXTURE_PROFILE=local-seed-obec-report-v1
+# Ontario 3rd-Party / Bonus-Codes bill entry (billing-on-third-party-playwright-checks.js).
+# Read-only: it opens the Ontario bill form for this appointment and switches the bill
+# type, but never submits a bill, so it seeds nothing and cleans nothing up. It asserts
+# billRegion travels on every self-navigation, which is what #3588 pinned -- an install
+# that has the `billregion` property set still routes correctly with the parameter
+# missing, so a status-only assertion would pass on a re-broken page.
+export BILLING_APPOINTMENT_NO=11 BILLING_DEMOGRAPHIC_NO=1 BILLING_PROVIDER_NO=999998
+export BILLING_APPOINTMENT_DATE=2024-04-16 BILLING_START_TIME=12:00:00
 # Rich Text Letter print/PDF check (fixture c above); omit to skip only its template step.
 export RTL_TEMPLATE_NAME=MissedAppointment.rtl
 # Rx signature-stamp fax check (rx-fax-signature-stamp-playwright-checks.js). It writes and then
--- a/package.json
+++ b/package.json
@@ -55,7 +55,8 @@
     "test:report-demographic-navigation-playwright": "node scripts/report-demographic-navigation-playwright-checks.js",
     "test:flowsheet-admin-playwright": "node scripts/flowsheet-admin-playwright-checks.js",
     "test:select-forms-panel-playwright": "node scripts/select-forms-panel-playwright-checks.js",
-    "test:flu-billing-report-playwright": "node scripts/flu-billing-report-playwright-checks.js"
+    "test:flu-billing-report-playwright": "node scripts/flu-billing-report-playwright-checks.js",
+    "test:billing-on-third-party-playwright": "node scripts/billing-on-third-party-playwright-checks.js"
   },
   "dependencies": {
     "@docusaurus/core": "^3.9.2",
--- a/scripts/billing-on-third-party-playwright-checks.js
+++ b/scripts/billing-on-third-party-playwright-checks.js
@@ -28,17 +28,25 @@
  * navigation query string WITHOUT billRegion, and Billing2Action's fall-back
  * read a holder nothing populated, so the router answered "BC" and handed an
  * Ontario request to billingBC.jsp -- which queries the BC-only billingvisit
- * table. The bill entry was lost with the 500. The same release also prefixed
- * the Ontario unbilled "Bill" links with the context path (they 404'd at the
- * host root) and put billRegion=ON on them.
+ * table. The bill entry was lost with the 500. The same release prefixed the
+ * Ontario unbilled "Bill" links with the context path (they 404'd at the host
+ * root) and put billRegion=ON on them.
  *
  * Why the region assertion and not just "no 500": the fall-back is a property,
  * `billregion`, and an install that HAS it set (the packaged Ontario install
  * does) still routes correctly with the query parameter missing. Asserting only
- * the status code would therefore pass on a re-broken page. This check asserts
- * the thing the fix actually pins -- billRegion travelling on every self
- * navigation -- and treats the rendered Ontario form and the absence of an
- * error page as the user-visible corroboration.
+ * the status code would therefore pass on a re-broken page.
+ *
+ * Page identity is asserted from the Ontario bill-type CODES (ODP/BON), not
+ * from the presence of a control named xml_billtype: billingBC.jsp carries an
+ * xml_billtype field of its own, so that control cannot tell the two forms
+ * apart. The codes can -- BC offers ICBC/WCB/Pri. Codes are also locale
+ * independent, unlike the page title.
+ *
+ * Each bill-type switch must actually navigate. A swallowed navigation would
+ * leave the browser on the entry URL, which already carries billRegion=ON and
+ * already renders the Ontario form, so every assertion would pass without the
+ * branch having run at all.
  *
  * READ-ONLY: it switches the bill type and reads links. It never submits a
  * bill, so it seeds nothing and has nothing to clean up.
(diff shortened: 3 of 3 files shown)


yingbull:
Addressed the review in `645d336e` — one comment rather than sixteen replies. The findings were right: the first cut could pass without proving anything.

**Page identity (sourcery, Codex).** Correct, and I checked before changing it: `billingBC.jsp` does carry an `xml_billtype` field (`document.BillingCreateBillingForm.xml_billtype`, `jQuery("#xml_billtype")`), so the control's presence cannot tell the forms apart and a rendered misroute would have satisfied the assertion meant to catch it. Identity now comes from the Ontario bill-type **codes** (`ODP`, `BON`); BC offers `ICBC`/`WCB`/`Pri`. Codes are locale-independent, unlike the title. The `billingvisit` probe is dropped — as noted, it is a server-side table name that never appears in rendered markup.

**Navigation must complete (sourcery, Codex).** Correct and the most serious one. The swallowed `waitForNavigation` left the browser on the entry URL, which already carries `billRegion=ON` and already renders the Ontario form — so a branch that stopped navigating passed everything after a 60s timeout. Each option now asserts navigation completed *and* that the resulting URL carries the `xml_billtype` code that was selected.

**The unbilled "Bill" links (Codex, sourcery, Copilot).** This was a guaranteed false SKIP, not a dataset gap — worse than a failure because it looked benign. Three causes, all fixed:
1. the report needs `reportAction=unbilled`, otherwise the assembler returns an empty model;
2. the anchors are `href="#"` with the URL in their `onclick="popupPage(...)"` handler;
3. that URL is written through `<carlos:encode context="javaScriptAttribute">` in **double** quotes with separators escaped (`&` → `\x26`), so a single-quote, literal-`&` matcher could never match.

`ViewBillingReportControl` now genuinely asserts both the context-path prefix and `billRegion=ON`. `ViewBillingONNewReport` renders no such link with these parameters and still skips — honestly this time.

On `reportAction`: I initially pushed back that the valid values were `billed`/`unsettled`/`billob`/`flu`. **That was my error** — I read the `@param` JavaDoc rather than the code. `case "unbilled":` is a real dispatch branch calling `loadUnbilledRows(...)`. Codex was right. Worth a separate one-line follow-up: that JavaDoc on `BillingReportFragmentViewModelAssembler.assemble` documents four of the five cases and omits `unbilled` — the one that produces the very links #3588 fixed. I've left it alone to keep this PR test-only.

**TLS (Codex P1, Copilot).** Correct contradiction. `ignoreHTTPSErrors` is now tied to a loopback host, so a target opted in via `ALLOW_NON_LOCAL_BASE_URL` must prove its certificate before this script submits administrator credentials to it.

**Copilot's three smaller ones** are all fixed by reusing the shared `eform-local-playwright-utils` helpers instead of duplicating them: `validateBaseUrl` normalises a root context to `""` (so the prefix check cannot look for `//billing?`), `buildArtifactPath` validates the artifact directory before creating it (the raw `fs.mkdirSync` on the unvalidated env value is gone), and the unused `path` import is removed.

**Semgrep `playwright-goto-injection` ×3.** Routed through the shared `gotoApp`, which carries the reviewed `nosemgrep` justification and rejects non-root-relative paths. The record pointers are also shape-checked (digits, `YYYY-MM-DD`, `HH:MM:SS`) before reaching a URL, so no unvalidated env value is interpolated.

**Re-verified both directions after the rewrite**, against the packaged Ubuntu 26.04 deb install:

| Deployed `billingON.jsp` | Result |
| --- | --- |
| As shipped | **30/30** |
| `billRegion=ON` reverted out of `__commonQs` | **24/30** — exactly the six region assertions fail |

Navigation, form-identity, error-page and report-link assertions all still pass while reverted, which is the point: only the assertion tied to the fix moves.

---

yingbull:
Second round addressed in `60281b11`. Six findings still applied to the post-rewrite file; all are fixed.

**PIN field (Copilot)** — the most consequential. `login()` filled `#pin` unconditionally, but `login/index.jsp:657` renders that input only inside `<% if (MfaManager.isOscarLegacyPinEnabled()) { %>`. On an install with legacy PIN disabled the fill throws and the check never runs at all. Now conditional on the field being present. Note the unconditional pattern also exists in sibling scripts (`assign-role`, others) — I have not touched those here, but they will fail the same way on such an install.

**`some()` in the coverage assertion (cubic P2)** — correct. A code disappearing from the control (`OCF`, `STD`…) still passed while the loop below silently stopped exercising that branch. It now requires *every* expected code and names the missing ones.

**Report route errors reported as SKIP (Copilot, cubic P2)** — correct. A 404/500 left the previous page loaded and fell through to the empty-list branch, so a broken route looked like a dataset gap. The response status is now checked; anything `>= 400` fails before a skip is possible.

**Dynamic `RegExp` (Semgrep `detect-non-literal-regexp`)** — the code came from a constant in this file, so there was no real ReDoS exposure, but a dynamic expression buys nothing here. The URL is parsed and `xml_billtype` compared directly; the `billRegion` assertion moved to the same parsed form.

**Identifiers in diagnostics (cubic P2 ×2)** — correct, and it matters here specifically: `CLAUDE.md` classes `demographic_no`, appointment IDs and provider numbers as PHI-correlating operational identifiers. Failure details were printing whole URLs carrying `appointment_no`, `demographic_no`, `demographic_name`, `providerview` and `user_no` into both console output and the results JSON. `redactUrl()` now keeps only what a routing failure needs — `billRegion`, `xml_billtype`, `curBillForm`, `billForm`, `reportAction` — and marks that it dropped the rest. The skip line no longer names the provider.

**Re-verified both directions**, because the load-bearing assertion changed implementation:

| Deployed `billingON.jsp` | Result |
| --- | --- |
| As shipped | **30/30** |
| `billRegion=ON` reverted out | **24/30** — exactly the six region assertions |

Redaction is visible in the failing output, which now reads:

```
FAIL 3rd party (PAT) keeps billRegion=ON on its navigation:
  url was /carlos/billing?curBillForm=PRI&xml_billtype=PAT (identifiers redacted)
```

Still diagnosable, no patient- or provider-correlating identifiers.

The earlier threads from the first round are resolved/outdated against `645d336e`; the ones above were the remainder that still applied to current code.

---

Issue #12, opened 2026-02-03
Hibernate 6 Prep: Migrate org.hibernate.Query imports to javax.persistence.Query


yingbull:
## Summary
Replace Hibernate-specific Query import with JPA-standard Query interface.

## Background
`org.hibernate.Query` has a different API in Hibernate 6. Using JPA's `javax.persistence.Query` provides forward compatibility.

## Scope
- **9 imports** across **9 files**

## Affected Files
- `IssueDAOImpl.java`
- `SecuserroleDaoImpl.java`
- `SecobjprivilegeDaoImpl.java`
- `SecProviderDaoImpl.java`
- `ProviderDaoImpl.java`
- `ProgramTeamDAOImpl.java`
- `ProgramFunctionalUserDAOImpl.java`
- `ProgramClientStatusDAOImpl.java`
- `CaseManagementNoteDAOImpl.java`

## Migration Pattern
```java
// Before
import org.hibernate.Query;

// After (works in H5 AND H6)
import javax.persistence.Query;
// or for typed queries:
import javax.persistence.TypedQuery;
```

## Verification
- [ ] All `org.hibernate.Query` imports replaced
- [ ] Code compiles with JPA Query interface
- [ ] Unit/integration tests pass

yingbull:
completed

PR #3622, opened 2026-09-06
fix(rx): record searched allergies, and alert on free-text ones when prescribing

yingbull:
## Description

Two defects from an alpha report against 2026.08, both reproduced and fixed against a deb install on Ubuntu 26.04 with the demonstration dataset.

### 1. Searching for an allergy and clicking a result did nothing

Reported console output, reproduced verbatim:

```
Uncaught ReferenceError: submitAddReaction is not defined
Uncaught TypeError: can't access property "indexOf", param is undefined
    sendSearchRequest .../rx/showAllergy?demographicNo=…
```

`ShowAllergies2.jsp` posts the search to `ChooseAllergy2.jsp` and lifts **only** its `#searchResultsContainer` element out of the response. That page's `<script>` block — which defines `submitAddReaction` — and its hidden `#addReactionForm` therefore never reach this DOM, while the result anchors' inline `onclick="submitAddReaction(...)"` resolves against `ShowAllergies2`. Hence the `ReferenceError`.

The second error was `ShowAllergies2`'s own delegated handler: it read the drugref id off `this.href`, but those anchors are `href="javascript:void(0)"` carrying the id, type and description in `data-` attributes, so it passed `undefined` into `sendSearchRequest`.

`ShowAllergies2.jsp` now defines `submitAddReaction` itself, routing into the AJAX reaction dialogue it has always driven, and the delegated handler is **removed** rather than left to fire twice alongside the `onclick`. Two related hardenings: the NKDA guard now decodes the parameter string before matching (`submitAddReaction` percent-encodes the name while the NKDA button passes it verbatim), and `sendSearchRequest` no longer dereferences a missing `param`.

The search POST was a clean `200` throughout — which is why the page looked healthy until a result was clicked, and how this shipped.

### 2. Rx of Amoxil to a patient with a Penicillin allergy did not alert

DrugRef's `get_allergy_warnings` dispatched only on the allergy's category and **silently skipped type 0** — the category CARLOS writes for every "Custom Allergy", and the one the demonstration dataset seeds its PENICILLINS allergies with (patients 12–18). It returned an empty warnings array: no alert, no error, and a response indistinguishable from "no allergy". A *typed* (AHFS class) penicillin allergy alerted correctly the whole time, which is what makes this easy to miss.

That half of the fix is in DrugRef (carlos-emr/drugref2026#12): a free-text allergen name is resolved against `cd_drug_search` by exact name and checked as whichever categories carry it; a name the reference does not know is reported as unchecked rather than as no allergy. This PR moves `debian/drugref.pin` onto it.

## Related Issues

Alpha bug report against 2026.08 (no issue number supplied).

Depends on carlos-emr/drugref2026#12 — **the CARLOS change alone fixes only defect 1**; the alert needs `carlos-emr-drugref` rebuilt at the new pin.

## Target Branch

`release/2026.08` — supported fix for the release the report was filed against, branched from and targeting that same line. No application version or SCM metadata is changed; `debian/changelog` gains one `2026.09.0~snapshot21` entry in the existing per-fix pattern.

`debian/drugref.pin` currently points at a **branch** commit on carlos-emr/drugref2026, not a master merge commit, because the two halves ship together. The pin file carries a maintainer note to re-point it once #12 merges.

## How Was This Tested?

Built both debs from this branch and installed them into a **fresh, disposable Ubuntu 26.04 systemd container** — demo dataset, nginx + ModSecurity/CRS on `:443`, DrugRef as a second webapp in the same Tomcat. The deb build fetched DrugRef at the new pin (`drugref source at 077788d…`), so the packaged artifact is what was tested, not a hot-patched class.

**Reproduced first.** Against the unfixed build the check reports both original errors, with the anchor and `sendSearchRequest` stack frames matching the tester's report.

**The reported scenario.** Prescribing Amoxil to demo patients 13/14/15 now renders `Allergy: PENICILLINS Reaction: Anaphylaxis`; before the fix the allergy probe returned `{"results":[]}` and no alert appeared.

**No false positives.** Azithromycin warns only on a MACROLIDES allergy, not the penicillin ones; `"No Known Drug Allergies"` and other names the reference has no row for are reported as unchecked rather than warned on.

**New regression check** — `scripts/allergy-rx-alert-playwright-checks.js`, run through `:443`, registered as `npm run test:allergy-rx-alert-playwright` and picked up by the `scripts/*-playwright-checks.js` loop in the deb-install runbook. It runs in three parts: record an allergen from the search results (typed), record a second through "Custom Allergy" (free text), then prescribe a drug in the *second* allergen's class and require both the probe JSON to carry that run's reaction marker and the alert table to be visible.

The two allergens are deliberately unrelated classes — penicillins for part 1, macrolides for part 2. Without that separation a regression in the free-text branch would be masked by the typed allergy part 1 just added, and the check would pass against the exact build the tester reported. **Verified to fail on each unfixed half separately and to pass with both in place:**

| build | result |
|---|---|
| unfixed JSP + unfixed DrugRef | fails part 1, reports both console errors |
| fixed JSP + unfixed DrugRef | fails part 3, empty warnings array |
| fixed JSP + fixed DrugRef | passes |

**No regressions on adjacent surfaces.** `drug-search-playwright-checks.js` and `echart-playwright-checks.js` pass against the same install, and `carlos-ctl check` is all-green (front door, WAF, DrugRef live lookup, 430 tables, 19 migrations).

## Screenshots

Not included — the change is a JS/JSP correctness fix; the visible outcome is the allergy alert row, whose exact rendered text is asserted by the new check and quoted above.

## Checklist

- [x] I preserved the target branch version/SCM metadata, or this is an explicit release-preparation PR

https://claude.ai/code/session_013pMK9W99ZHRuEoEL8pZTs9

## Summary by Sourcery

Fix allergy search selection and free-text allergy warning behavior in the prescription workflow.

Bug Fixes:
- Restore allergy recording from search results and ensure prescribing alerts for free-text allergies resolved by DrugRef.

Enhancements:
- Harden allergy search request handling and NKDA detection against encoded or missing parameters.

Build:
- Update the packaged DrugRef dependency pin and record the release changelog entry.

Documentation:
- Document the allergy prescription alert check and its deb-install validation requirements.

Tests:
- Add a Playwright regression check covering typed and free-text allergy recording and prescription alerts, including the reported penicillin scenario.


yingbull:
All five review findings verified and fixed in `5528708`. Two of them were pointing at hollow assertions in the new check, which is exactly the failure mode `deb-install-validation.md` warns about, so thank you for those.

**Finding 4 — the persistence assertion could pass vacuously.** Correct, and worse than the comment says: the check *chooses* PENICILLINS, and the demonstration dataset seeds PENICILLINS allergies on those very patients, so `listText.includes("PENICILLINS")` was satisfied by the pre-existing row whether or not "Add Allergy" saved anything. Every persistence and alert assertion now keys on the unique reaction text that run writes, which only that run's recording can produce.

**Finding 1 — the typed allergen was recorded and never alerted on.** Also correct: part 3 only ever prescribed a macrolide, so the penicillin half of the check stopped at persistence. Added a part 4 that prescribes `amoxil` against the typed allergy and requires *its* marker back — which is the scenario as the tester reported it. The two classes stay disjoint, so neither part can be satisfied by the other's allergy. The header env list drifted from the config too (`amoxil` vs `biaxin`); both are now documented with which class each belongs to.

**Finding 2 (P1) — certificate verification off-loopback.** Valid, and there is precedent I had missed: `billing-on-third-party-playwright-checks.js` gates `ignoreHTTPSErrors` on loopback with this exact reasoning. Adopted, and noted in the runbook. `ALLOW_NON_LOCAL_BASE_URL` should not silently also mean "accept any certificate" when the script is submitting clinician credentials.

**Finding 3 — flat results.** Valid; the check would have timed out under `allergies.flat_results=true`. Both layouts put the drugref id on the anchor itself, so it selects `a[data-id]` from `#searchResultsContainer` — which also skips the section toggles. Verified passing with that property set both `false` and `true`.

**Finding 5 — PHI in diagnostics.** Valid. The probe results and the alert row's text are the patient's own recorded reactions. Assertions now report counts and leave the detail to the screenshot; the failure block no longer serialises response bodies.

Re-verified on the packaged deb install, on a loopback `BASE_URL` through `:443`: all four parts pass, and the check still fails on each unfixed half separately — part 1 on the old JSP with the original `ReferenceError`, part 3 on the old DrugRef with an empty warnings array.

`debian/drugref.pin` also moves to `5b5ff3a`, which carries the DrugRef half of that review — including a class-level ATC allergen fix that was the same silent-miss shape as the bug this PR fixes.

---

yingbull:
Pin moved again, to `6bbeb3a` — a second DrugRef review round found three more ways an allergy could come back **unchecked but indistinguishable from safe**, plus one that could have invented an alert on every prescription (a blank ATC code prefixes every drug; 6,529 rows in the seeded dataset are one data change away from that). All fixed and re-verified; details on carlos-emr/drugref2026#12.

**One thing I deliberately did not fold into this PR**, so it does not get lost:

CARLOS renders no alert when DrugRef reports an allergy as *unchecked*. `RxShowAllergy2Action.getAllergyData` drops the `missing` list entirely, and its own `catch` writes `{"results":[]}` — the same shape as "this patient has no allergy to this drug". So even now that DrugRef distinguishes "checked and clear" from "not checked", a prescriber cannot tell them apart: an unresolvable allergen name, a DrugRef outage, or a failed query all render as silence next to the drug.

That is the same class of defect as the bug this PR fixes, one layer up. Fixing it means changing the action's JSON contract and `prescribe.jsp` to render something like "allergy check unavailable" beside the drug — a UI change on a surface this PR otherwise only reads from, and beyond what the tester reported. I would rather it be its own change with its own review than widen this one. Happy to open the issue, or the PR, on a nod.

The four-part browser check passes end to end against the packaged deb install on the current pin.

---

yingbull:
Copilot's HTTPS point was valid and is fixed in `3b655a4`. The header told operators to run through `:443` but nothing said so at run time, and the default `BASE_URL` is plain HTTP — so a run could quietly bypass nginx and ModSecurity/CRS while the output looked identical to a representative one.

Worth naming what that costs here specifically: this check posts free clinical prose — the reaction description — on two separate forms, which is the shape CRS has blocked on other surfaces in this repo. Now warns on stdout, mirroring `drug-search-playwright-checks.js`. Verified it fires on an HTTP `BASE_URL` and stays silent on the HTTPS run, which still passes all four parts.

The remaining cubic re-flag on drugref#12 (null AHFS number) was from a review of `b1e58237`, one commit before the fix; cubic has since marked it addressed in `6bbeb3a`, and I proved it with a temporary row carrying an AHFS description and a null number.

---

yingbull:
Copilot's RegExp finding is valid and fixed in `dd514f1`. It is also less hypothetical than the comment suggests — I scanned the seeded reference to check:

- **274 of the 75,113 allergen names throw** when interpolated raw. `PERFECTING DAY CREAM BROAD SPECTRUM SPF 30 PA +++` → `Invalid regular expression: Nothing to repeat`.
- Many more **silently mis-match**, because `.` matches any character: the pattern built from `TEARS NATURALE II 0.1%/0.3%` happily matches `TEARS NATURALE II 0X1%/0X3%`.

The default `PENICILLINS` is unaffected, so this was latent — but the entire point of `ALLERGY_ALLERGEN` is to aim the check at a different allergen, and it would have failed in the two least useful ways available: a crash, or a quietly wrong match that keeps reporting green.

Now escaped before the pattern is built. Verified across all 75,113 names: none throw, and each matches only itself (checked both bare and with the surrounding whitespace the anchor allows). The four-part check still passes end to end on the packaged deb install.

---

yingbull:
Valid, and the sharpest framing of it is cubic's own: part 4 would have been "checking a pairing nobody configured". Fixed in `c71019c`.

Part 4 prescribes `ALLERGY_TYPED_DRUG_TERM` and requires a warning for the allergen part 1 recorded. That assertion only means anything because the two are a deliberate pair — amoxil against PENICILLINS. The fallback that lets parts 1–3 run on a dataset lacking the requested name silently broke that pair, and the two possible outcomes were both bad: fail for a reason that is not a defect, or **pass by coincidence** (the fallback lands on some other penicillin) while testing something nobody chose. The second is the one that matters — this whole check exists because of assertions that pass without proving anything.

Part 4 now runs only when the search actually returned `ALLERGY_ALLERGEN`, and otherwise prints a `[skip]` line naming what got recorded instead — the pattern `eform-consultation-acceptance` already uses for its template probe. Parts 1–3 are untouched, so the fallback keeps doing its job.

Verified both paths:

```
# default configuration
recorded from search: PENICILLINS
biaxin warned on the free-text allergen
amoxil warned on the allergen recorded from the search results

# ALLERGY_ALLERGEN='ZZ NOT IN RESULTS'
[skip] part 4: the search for "penicillin" did not return "ZZ NOT IN RESULTS", so part 1
recorded "BENZATHINE BENZYLPENICILLIN" instead. ALLERGY_TYPED_DRUG_TERM ("amoxil") is paired
with "ZZ NOT IN RESULTS", not with that, so asserting a warning here would be checking a
pairing nobody configured. Parts 1-3 still ran.
recorded from search: BENZATHINE BENZYLPENICILLIN
biaxin warned on the free-text allergen
```

Both exit 0 against the packaged deb install.

---

yingbull:
Semgrep PRO's `detect-non-literal-regexp` (ReDoS) on the allergen matcher is resolved in `65217c2` — by removing the construct, not suppressing it.

On its own terms the finding was a false positive twice over: the argument was escaped to a literal, so the pattern could not backtrack, and the "attacker-controlled input" is an environment variable set by whoever is already running the script on their own machine. Per CLAUDE.md that would have been a legitimate candidate for a rule-specific `nosemgrep` with justification.

I did not do that, because the regex was the wrong tool anyway. The comparison I want is **equality** — and expressing equality as an anchored pattern is exactly what created the escaping problem Copilot flagged one round earlier. Comparing the trimmed result text directly is shorter, exact by construction, drops the `escapeRegExp` helper entirely, and leaves nothing to suppress. One fewer suppression in the tree is worth more than a defended one.

Verified all three behaviours on the packaged deb install:

- default configuration → all four parts run and pass
- `ALLERGY_ALLERGEN='ZZ NOT IN RESULTS'` → part 4 skips with its note, parts 1–3 pass
- `ALLERGY_ALLERGEN='PENICILLIN'` (a strict substring of several real results) → correctly does **not** match, so equality semantics are intact rather than having quietly become "contains"

---

yingbull:
@coderabbitai full review

---

yingbull:
@coderabbitai review

---

yingbull:
@phc007 for awareness

yingbull:
@copilot+claude-opus-5:high resolve the merge conflicts in this pull request

PR #3677, opened 2026-09-14
fix(edoc): remove the Add HTML button from the document manager

1984Doc:
The Add HTML tab let a user compose a blank HTML document from scratch, which was rarely used and easy to confuse with Add Link (both are backed by the same AddEditHtml2Action, differing only by mode). Removing it only drops the ability to start a brand-new blank HTML document from this UI.

Left untouched, since they're shared with other flows:
- AddEditHtml2Action's addHtml/addLink mode branching (Add Link still uses it)
- ViewAddEditHtml / addedithtmldocument.jsp / addEditHtml route, which is also how an existing HTML or Link document is edited (documentReport.jsp, documentBrowser.jsp both link here via editDocumentNo)
- eForm's own HTML-document saving (EFormUtil), which calls the underlying EDoc HTML-content-type mechanism directly, never through this JSP

Verified live: Add Document and Add Link still render and expand correctly, no console errors, and the edit-existing-document code path was not modified.

## Checklist

- [ ] My commits are signed off for the [DCO](https://developercertificate.org/) (`git commit -s`)
- [ ] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format, or I've written clear commit messages and will use the format next time
- [ ] I have not included any patient data (PHI) in this PR
- [ ] I have added tests for new functionality, or this change doesn't need new tests
- [ ] I have read the [contributing guide](https://github.com/carlos-emr/carlos/blob/develop/CONTRIBUTING.md)
- [ ] I selected the target branch according to the [release process](https://github.com/carlos-emr/carlos/blob/develop/docs/release-process.md)
- [ ] I preserved the target branch version/SCM metadata, or this is an explicit release-preparation PR
- [ ] I did not modify a Flyway migration that exists in a published release tag

## Summary by Sourcery

Enhancements:
- Remove the Add HTML option from the document manager while retaining existing HTML document editing and Add Link functionality.

---
## Summary by cubic
Removes the Add HTML button from the document manager so users can no longer start a new blank HTML document from that screen. The button was rarely used and easy to confuse with Add Link, which is unchanged. Editing existing HTML and Link documents still uses the existing `ViewAddEditHtml` route, and `eForm`'s HTML-saving flow is unaffected. Verified manually that Add Document and Add Link still render and expand without console errors.

<sup>Written for commit ce664a7e3391a22ae310b71e74d54ef3c39f6052. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/carlos-emr/carlos/pull/3677?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


1984Doc:
@claude review

yingbull:
@phc007 thoughts on this?

phc007:
I hate removing features without thinking about the diversity of end users.

That said  I note that this is a feature that has never been used by our
group.  I also note that you can save an URL which is preserved.

All in all it's probably ok to drop (an easy reversion in any case)

The PR won't simplify maintenance, but it does clean the UX marginally

1984Doc:
Agreed.  I did Claude and on the back end it remains for eform
functionality.  It's just dropping the link as it is confusing for the end
user too.  It's not removing the feature all together, only the link.

yingbull (approved):

Issue #3728, opened 2026-09-17
[Bug] Schedule page polls /SystemMessage and /FacilityMessage, which require `_admin` write, so every non-admin provider gets two 403s per load


yingbull:
## Summary

`WEB-INF/jsp/provider/appointmentprovideradminday.jsp` fetches two banners on every schedule load:

```
:1558  jQuery.get(".../SystemMessage?method=view", "html", ...)
:1561  jQuery.get(".../FacilityMessage?method=view", "html", ...)
```

`SystemMessage2Action` (`www/SystemMessage2Action.java:59`) throws `SecurityException("missing required sec object (_admin)")` unless the caller holds `_admin` **write**. The facility-message action behaves the same way.

## Observed (fresh 2026.08.0-alpha12 deb)

A throwaway provider with only the `doctor` role logs in fine, but the schedule page records:

```
403 GET /FacilityMessage?method=view&html
403 GET /SystemMessage?method=view&html
```

on every load. `carlosdoc` (doctor + admin) does not see them, which is why the suites never noticed.

## Why it matters

Two failed requests per schedule load for every clinician, plus two `SecurityException` WARN lines with incident ids in the Tomcat log per load (the exception interceptor added by #3623 logs each one), which will drown real security warnings on a busy clinic.

## Suggested fix

Viewing a system/facility message is not an admin write. Either gate the `view` method on read (`_admin` "r", or a dedicated `_msg`-style object), or have the JSP only poll when the provider holds the privilege (`<security:oscarSec>` around the two fetches).

Found while verifying #3701.

yingbull:
Verified fixed on `release/2026.08` — closing (PR #3745).

A new `www/MessageBannerAction.java` base class special-cases `"view"` to require only a logged-in session (`if (loggedInInfo == null) throw ...`), with `_admin` write still enforced for edit/save/list.

Both endpoints you named are covered by that one base class: `SystemMessage2Action` extends it, and `FacilityMessage` is not a separate action — `struts-admin.xml:56` maps `name="FacilityMessage"` → `OrganizationMessage2Action`, which also extends it. So both `jQuery.get` polls at `appointmentprovideradminday.jsp:1557` and `:1560` stop 403-ing for non-admin providers, and the paired `SecurityException` WARN lines stop filling the log.

Note: the fix is on the release line; `MessageBannerAction` does not exist on `develop`, where both actions still check `_admin`/`w` before dispatch — pending the forward-merge tracked in #3726.

---
[...]

PR #3745, opened 2026-09-18
fix(admin): let any authenticated provider read the schedule message banners

yingbull:
## Description

The provider schedule (`WEB-INF/jsp/provider/appointmentprovideradminday.jsp:1557,1560`) fetches `/SystemMessage?method=view` and `/FacilityMessage?method=view` on every load. Both actions gated *every* method — including the read-only `view` — behind `_admin` **write**, so any provider without admin rights got:

- two 403s per schedule load, and
- two `SecurityException` WARN lines with incident ids in the Tomcat log per load (the exception interceptor from #3623), which would drown real security warnings on a busy clinic.

They also never saw the broadcast notices the banners exist to deliver. `carlosdoc` is doctor + admin, which is why the suites never caught it.

System and facility messages are administrator-authored notices addressed to all staff and hold no PHI, so `view` is now authorized on an authenticated session alone:

- `?method=view` requires a `LoggedInInfo` in session and nothing more. `LoginFilter` (mapped at `/*`) is the canonical session gate; the null check is defence-in-depth for direct invocation and filter reordering.
- The management methods (`list` / `edit` / `save`) keep their `_admin` write gate, unchanged.
- The facility `view` still scopes itself to the session's current facility and the caller's own program domain.

Both actions previously dispatched `view` *after* the privilege check; the dispatch is now split so the read path is reached without consulting `SecurityInfoManager` at all.

That rule lives once in a new `MessageBannerAction` base which both actions extend. Restating it in two near-identical `execute()` methods is what failed SonarCloud's duplication gate at 10.2%; extracting it took the PR to 0.0%.

## Related Issues

Fixes #3728

## Target Branch

Targets `release/2026.08`: the bug was reported against a fresh `2026.08.0-alpha12` deb, which makes 2026.08 the oldest affected supported release line. Per `docs/release-process.md`, maintainers forward-merge from here rather than a duplicate PR against `develop`. No version/SCM metadata was touched and no Flyway migration was modified.

## How Was This Tested?

> **Corrected.** An earlier version of this section named `mvn test -Dtest='SystemMessage2ActionTest,OrganizationMessage2ActionTest'` and reported 8 tests. That command was wrong: `OrganizationMessage2ActionTest` does not exist, and `SystemMessage2ActionTest` is the legacy class listed in `src/test/resources/surefire-unmatched-baseline.txt`, which the pom's Surefire `<includes>` never select — so a reviewer running it would not have executed the added coverage at all. Raised in review; the accurate command and count are below.

```bash
mvn test -Dtest='SystemMessage2ActionSecurityUnitTest,OrganizationMessage2ActionUnitTest'
```

**10/10**, 5 cases in each class, all inherited from `MessageBannerAuthorizationUnitTestBase`:

- the banner renders with `_admin` write **and** read both denied, asserting `hasPrivilege` is never consulted on that path;
- an unauthenticated `view` is rejected with no DAO interaction;
- `save`, `edit` and the default `list` are each still rejected without `_admin` write, with `verifyNoInteractions` on the management collaborator.

The `edit` case was added in `4680792dd3` after review pointed out that the contract covered `save` and `list` but not the third administrator-only route.

Also run: `MutatorActionGetRejectionContractUnitTest` — **41/41**, confirming the extraction did not disturb the GET/HEAD rejection manifest.

**Why the new class names.** Coverage added to `SystemMessage2ActionTest` would never have run in CI, for the Surefire reason above. It could not simply be renamed either: two of its pre-existing `edit()` cases error with `getContainer() is null`, reproducible by stashing this branch and running against unmodified `release/2026.08`, so enabling the class would turn CI red for an unrelated defect. The new coverage therefore lives in `SystemMessage2ActionSecurityUnitTest`, which CI does select. Those two pre-existing errors are neither caused nor fixed here and are worth a separate issue.

**Not run:** Tomcat/Playwright verification of this specific change. The behavioural check a reviewer would want is a non-admin provider (doctor role only) loading the schedule and seeing 200s instead of 403s on both banner fetches.

## Checklist

- [ ] My commits are signed off for the [DCO](https://developercertificate.org/) (`git commit -s`) — **see below**
- [x] My commits follow [Conventional Commits](https://www.conventionalcommits.org/) format
- [x] I have added tests for new functionality
- [x] I preserved the target branch version/SCM metadata

> **DCO is not satisfied and the check cannot be trusted on this PR.** `5e8de911` has no `Signed-off-by`; `4680792dd3` does. The remedy is `git rebase --signoff origin/release/2026.08` followed by a force-with-lease push, which a maintainer must run — both commands are on the deny list `CLAUDE.md` sets for me. Note also that this PR's `DCO` status showed a **pass at one point that no human granted**: quoting the manual-confirmation phrase inside an explanatory comment triggered it, because `dco-check.yml` matches with a plain `contains()` over the raw comment body. Details in the thread.

https://claude.ai/code/session_01JVRt9rfEaFYnHm34oXQuRC


yingbull:
**`DCO Sign-Off` is red on `5e8de911` and it is mine: that commit has no `Signed-off-by` trailer.**

The other three commits on this branch have one. I dropped it on this one when I committed the SonarCloud duplication fix. No excuse — it's a slip in the commit, not in the change.

The workflow's own failure comment never made it onto the PR: after writing the `failure` status it hit `curl: (22) The requested URL returned error: 403` posting the comment, so the run ends with a bare red check and no explanation. That's why this comment exists.

```
##[error]The following commit(s) are missing a DCO sign-off:
  - 5e8de91 refactor(admin): hold the banner authorization rule in one action base
```

**What blocks me fixing it.** `.github/workflows/dco-check.yml` names the preferred fix as an amend with `-s` followed by a force-with-lease push. Both are on the deny list `CLAUDE.md` sets for me — amend under history rewriting, force push under force push. There is no way to add a trailer to an already-pushed commit without one of them, and a follow-up commit does not help because the check walks every commit in the PR. So I am not able to push the fix myself.

**Two ways to clear it**, both needing a human: run that amend and force-with-lease (the tree is correct as it stands; only the trailer is missing), or post the documented maintainer confirmation phrase for this revision.

I have deliberately not issued that confirmation myself. It is an attestation reserved for a maintainer with write permission, and self-issuing it to clear my own slip is not mine to do.

> **Edited.** This comment originally quoted the confirmation phrase verbatim, to show what the second option looks like. That was a mistake: `dco-check.yml` triggers on `contains(github.event.comment.body, ...)`, which is a plain substring test with no awareness of code fences, so quoting the phrase *was* issuing it. Posting this comment set the DCO status to "manually confirmed by yingbull" four seconds later. I have removed the phrase and explained the accidental confirmation in a follow-up comment below. No human made that attestation.

The change itself is unaffected and verified locally: `SystemMessage2ActionSecurityUnitTest` and `OrganizationMessage2ActionUnitTest` 8/8, and the mutator GET/HEAD rejection contract 41/41. I'll keep the PR watched.

---

yingbull:
## The DCO status on this PR is a false pass. Treat it as not given.

The `DCO` commit status currently reads **"DCO manually confirmed by yingbull"**. **No human made that attestation.** I triggered it by accident, and I want it on the record before anyone merges on the strength of it.

**What happened**

My previous comment explained that the maintainer confirmation was one of two ways to clear the red DCO check, and — to show what it looks like — quoted the phrase verbatim inside a fenced code block, immediately under a sentence saying I was deliberately *not* issuing it.

`.github/workflows/dco-check.yml` gates that path on:

```yaml
(github.event_name == 'issue_comment' &&
 github.event.issue.pull_request &&
 (contains(github.event.comment.body, '<the phrase>') || contains(..., '<lowercased>')) &&
 contains(fromJson('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association))
```

`contains()` is a plain substring test over the raw comment body. It has no notion of code fences, quoting, or negation. I post as `yingbull`, a MEMBER. So the quote satisfied every condition.

The timing is unambiguous:

| | |
|---|---|
| comment `5725789858` created | `05:51:17Z` |
| DCO run `35312447972` created | `05:51:21Z` |
| step "Check for manual DCO confirmation comment" | success |
| step "Set success status" | **skipped** — it did not pass on trailers |
| resulting status | `DCO manually confirmed by yingbull` |

No other comment on this PR contains the phrase.

**What I've done**

I edited the offending comment to remove the verbatim phrase and noted the edit inline. The workflow triggers on `types: [created]`, so the edit does not re-fire it — but it also does not retract the status that was already written, and I have no way to revoke a commit status.

**What is actually still owed**

`5e8de911` still has no `Signed-off-by` trailer. The underlying problem is unchanged; only the check is lying about it. The real fix remains an amend with `-s` plus a force-with-lease push, which needs a human — or a maintainer deciding, knowingly, to confirm.

**Worth hardening regardless.** Any reviewer who quotes the phrase while discussing DCO — in a comment, a checklist, a runbook, a reply — silently issues a confirmation on whatever PR they are looking at. Requiring the phrase to be the comment's first line, or rejecting it when fenced, would close that. I have not touched `.github/` here; flagging it for a maintainer.

---

yingbull:
Addressed the review finding on `5e8de911` in `4680792dd3`.

The point was right: `MessageBannerAuthorizationUnitTestBase` asserted that `save` and the default `list` are refused without `_admin` write, but not `edit` — the third administrator-only method on the dispatch this PR extracted. An unauthorised `?method=edit` could have regressed without a failing test. Both subclasses inherit the new case, so it covers the system and facility banners alike: 5 cases each, **10/10**.

**One consequence for the DCO fix.** The branch now has two commits to consider, not one: `5e8de911` still has no `Signed-off-by`, and `4680792dd3` has one. So the remedy is no longer a tip amend — it is the workflow's other documented form:

```bash
git rebase --signoff origin/release/2026.08
git push --force-with-lease
```

That signs every commit on the branch in one pass. Still yours to run; both commands are on the deny list `CLAUDE.md` sets for me.

And repeating the warning from further up this thread so it does not get lost: **the `DCO` status on this PR currently shows a pass that no human granted.** I caused it by quoting the confirmation phrase in an explanatory comment, which `dco-check.yml` matches with a plain `contains()` over the raw comment body. Please do not read that green check as the sign-off being satisfied.

---

yingbull:
Second review finding addressed — the PR description, not the code.

It was right, and it was the worst kind of stale documentation: the "How Was This Tested?" section named

```
mvn test -Dtest='SystemMessage2ActionTest,OrganizationMessage2ActionTest'
```

which runs **neither** of the classes that actually carry this coverage. `OrganizationMessage2ActionTest` does not exist, and `SystemMessage2ActionTest` is the legacy class in `surefire-unmatched-baseline.txt` that Surefire never selects. A reviewer following that command would have concluded the change was verified while executing none of it. The count was stale too — 8, when it is now 10 after the `edit` case.

Description corrected: accurate command, 10/10 across `SystemMessage2ActionSecurityUnitTest` and `OrganizationMessage2ActionUnitTest` (5 each), plus the `MutatorActionGetRejectionContractUnitTest` 41/41 run, and an explicit note on why the new class names exist rather than adding to the legacy one.

I also unticked the DCO box on the checklist. It was ticked, and it is not true: `5e8de911` still has no `Signed-off-by`. Leaving it ticked while the check fails would have been its own small piece of misinformation.

CI on `4680792dd3` is **39/40** — DCO is the only failure, correctly naming `5e8de91` alone.

---

PR #2744, opened 2026-06-08
SMS backend foundation with consent and recovery safeguards

Ben-Heerema:
## Summary

Related to #2741. Adds a provider-neutral SMS backend foundation: system-of-record persistence, direct/queued dispatch, database-coordinated rate limiting, stale-send recovery, audited body access, and authenticated-provider webhook processing boundaries. Normal clinical sends remain blocked by the deferred consent service; only explicitly enabled `SYSTEM_TEST` messages can pass. Real provider adapters and public UI/API endpoints are not implemented.

The existing branch was rebuilt on current `develop`, replaying only the original 26 SMS commits with authorship preserved and DCO sign-offs. The final PR diff contains 81 SMS-related files plus three existing test corrections exposed by the full suite. The SMS files include the active Flyway migration, Spring model registration and developer documentation; no unrelated historical changes were imported.

## Review fixes

- Consent is evaluated before persistence and committed with initial queue state. The worker rechecks consent before dispatch. Denied attempts discard the full body while retaining its audit fingerprint.
- Send/worker entry points suspend caller transactions. The send claim is committed before network I/O, so a caller rollback cannot erase the record of an external send.
- Provider exceptions, null results and ambiguous outcomes remain `SENDING` for status reconciliation. They are not blindly retried. A direct response reflects the persisted result when a delivery webhook wins the update race.
- An explicit invalid provider configuration blocks SMS rather than reporting simulated STUB success. The adapter contract requires a client reference; the overload that discarded it was removed.
- Callbacks cannot cross provider boundaries, mutate an inbound message, supply conflicting identifiers, introduce internal queue states, or omit a delivery status. Oversized opaque identifiers are rejected instead of truncated; database keys compare them case-sensitively.
- Record diagnostics redact bodies, phone numbers, metadata and provider free text. Default failure logging uses the transaction identifier without patient identifiers.
- Static-analysis follow-up separates fixture construction from exception assertions, asserts the redaction sample count, removes redundant STUB branches and simplifies scheduler bounds and metadata filtering.
- The obsolete update-script/oscardata seed path was replaced with `V1.0.25__add_sms_system_of_record.sql` in the active common Flyway migration set. Both provincial migration sequences have unique versions.

## Naming and architecture

[...]

Issue #2741, opened 2026-06-08
[STILL IN PROGRESS] Plan CARLOS two-way SMS backend with consent-aware provider-neutral architecture


Ben-Heerema:
## Summary

CARLOS needs a two-way SMS backend plan before implementation. The backend should support outbound and inbound SMS, provider-neutral integration, message history, delivery tracking, and consent-aware sending.

This ticket is still in progress and should be refined alongside the consent work in #2674 and #2692.

## Related Work

- #2674 - CARLOS Consent Handling Suggestions
- #2692 - Draft: Consent Patient Master File reference mocks
- #2404 - Decide what patient consent information CARLOS needs for SMS/MMS messaging
- #2443 - SMS Mocks For Discussion

## Current Codebase References

- Ontario billing architecture conventions: `docs/billing-ontario-module.md`
- Layer naming policy: `docs/architecture/layer-names.md`
- Current consent models/DAOs:
  - `Consent`
  - `ConsentType`
  - `ConsentDao`
  - `ConsentTypeDao`
  - `PatientConsentManager`
  - `PatientConsentManagerImpl`
- Current email communication patterns:
  - `EmailComposeManager`
  - `EmailConfig`
  - `EmailLog`
  - `EmailLogDao`
- Existing deprecated appointment reminder email flow:
  - `AppointmentMailer`

## Proposed Direction

Create a new SMS module using the Ontario billing naming/layer conventions.

Package:

```text
io.github.carlos_emr.carlos.sms
```

Suggested subpackages:

```text
web/
assembler/
viewmodel/
command/
dto/
service/
validator/
support/
```

Use `Sms` casing in Java identifiers, not `SMS`.

## Proposed Core Entities

Add shared entities under `commn.model` and matching DAOs under `commn.dao`:

- `SmsConfig`
- `SmsMessage`
- `SmsConversation`
- `SmsDeliveryEvent`

Avoid creating a separate SMS consent audit table unless #2674 does not provide usable `ConsentEvent`, `ConsentTypeVersion`, and `ConsentEvidence` support.

## Consent Requirements

SMS must integrate with the consent model from #2674:

- `Consent` remains the fast current-state check.
- `ConsentTypeVersion` records exact wording/version.
- `ConsentEvent` records grant, decline, withdraw, clear, reconfirm, and STOP/opt-out events.
- `ConsentEvidence` records proof such as verbal confirmation, paper form, portal confirmation, import, or patient reply.

SMS-specific behavior should include:

[...]

Issue #2674, opened 2026-06-03
CARLOS Consent Handling Suggestions


Ben-Heerema:
## Background/Current Set up
CARLOS currently uses `ConsentType` to define consent categories (E.g. email consent, research study consent, etc.) and `Consent` to store patient-level consent state. This works for current yes/no checks, but it is weak for audit history. Consent changes overwrite the same row, and mutable consent wording can make it unclear what a patient actually agreed to at the time as wording can be changed without reasking for consent.

## Diagram Legend
- `PK` means primary key: the field that uniquely identifies one row in that table.
- `FK` means foreign key: a field that points to a row in another table.

Example: `Consent.consent_type_id` is an `FK` because it points to `ConsentType.id`.

## Current Setup Diagram

```mermaid
erDiagram
    CONSENT_TYPE ||--o{ CONSENT : "classifies"

    CONSENT_TYPE {
        int id PK
        string type
        string name
        string description
        string providerNo
        boolean remoteEnabled
        boolean active
    }

    CONSENT {
        int id PK
        int demographic_no
        int consent_type_id FK
        boolean explicit
        boolean optout
        string last_entered_by
        datetime consent_date
        datetime optout_date
        datetime edit_date
        boolean deleted
    }
```

```mermaid
flowchart TD
    UI[Demographic consent UI / API caller] --> Manager[PatientConsentManagerImpl]
    Manager --> TypeDao[ConsentTypeDao]
    Manager --> ConsentDao[ConsentDao]
    TypeDao --> CT[(consentType table)]
    ConsentDao --> C[(Consent table)]

    Manager --> Check{Consent row exists?}
    Check -->|No, opt in| Insert[Create Consent row]
    Check -->|Yes| Update[Update same Consent row]
    Check -->|Opt out| OptOut[Set optout=true and optout_date]
    Check -->|Clear| Clear[Set deleted=true]

    Insert --> C
    Update --> C
    OptOut --> C
    Clear --> C

    C --> CurrentCheck[hasPatientConsented]
    CurrentCheck --> Rule[Returns true when deleted=false and optout=false]
```

Current limitation: `Consent` is doing two jobs at once. It is both the current state used for fast checks and the closest thing to an audit record. Because the same row is updated, prior states are not preserved in a structured way.

## Goal
Add a consent model that preserves a clear audit trail while keeping existing current-state consent checks working and fast.

## Proposed Design
Keep `ConsentType` as the consent category.

Add:

```text
[...]

PR #3763, opened 2026-09-18
feat(sms): gate outbound SMS on a dedicated patient consent type

Ben-Heerema:
refs #2674, refs #2741

## Problem

The SMS backend merged in #2744 ships with `DeferredSmsConsentService`, a placeholder that blocks every patient message until the consent audit model in #2674 exists. #2674 has not moved since June, so nothing can be sent and the rest of the SMS work (adapters, send action, reminders) has nothing real to test against.

## What changed

- **Real consent check on the existing tables.** `CarlosSmsConsentService` reads the patient's current `Consent` row for the consent type named by a new `sms_communication` property (same mechanism as `email_communication`). Only a current opt-in permits a send:

  | Situation | Row status | Reason code |
  | --- | --- | --- |
  | Property unset, or consent type missing/inactive | `CONSENT_BLOCKED` | `SMS_CONSENT_NOT_CONFIGURED` |
  | No consent record, or record deleted | `CONSENT_BLOCKED` | `SMS_CONSENT_UNKNOWN` |
  | Patient opted out | `OPTOUT_BLOCKED` | `SMS_CONSENT_OPTED_OUT` |
  | Patient consented | proceeds | none |

- **Dedicated consent type.** Flyway `V1.0.29` seeds `sms_communication_consent` rather than reusing `electronic_communication_consent`, whose wording never mentions text messages. Existing patients start blocked until SMS consent is recorded. No new UI is needed: the patient record's consent section already lists every active consent type.
- **Consent audit snapshot.** Each `sms_transaction` row records `consent_status`, `consent_id` and `consent_last_update_date`, matching what `emailLog` records for email.
- **Worker hardening.** If the dispatch-time consent recheck throws, the worker now releases the claim back to `QUEUED` without consuming an attempt. Previously the exception escaped, aborted the batch, and left a never-sent row `SENDING` for stale recovery to treat as an uncertain send.
- **Duplicate consent rows fail safe.** `Consent` has no unique key on patient and consent type, and the single-row DAO lookup picks one live row arbitrarily. The check reads all of the patient's live records for the SMS type: any opt-out blocks the send, otherwise the most recently edited opt-in is recorded.
- The consent record is read through `ConsentDao`, not `PatientConsentManager`, because the manager's lookup requires a `LoggedInInfo` and the queue worker has no session.
- `docs/architecture/sms-backend.md` documents the decision table, the snapshot semantics and what is still deferred.

## Needs a human decision before merge

The seeded consent wording is a first draft and the migration cannot be edited once released:

> This patient has consented to receive text messages (SMS) from the clinic, including appointment reminders and administrative notices. Text messages are not encrypted and may be visible on a locked screen. The patient may withdraw consent at any time.

## Deliberately not in this PR

Consent scoped to a phone number or message type, STOP-reply opt-out, and the #2674 consent history tables.

## How was this tested?

- `mvn '-Dtest=**/sms/**/*Test' test`: 196 tests in 33 classes, 0 failures. New: 13 unit tests for the consent decisions (including conflicting duplicate records), 5 for the resolver, 5 integration tests against the real `property` / `consentType` / `Consent` tables (H2) including the snapshot round-trip, plus DTO, entity, recorder and worker tests.
- The worker regression test was confirmed to fail with the fix reverted (`IllegalStateException` escapes `processDueMessages`) and pass with it restored.
- `V1.0.29` was run twice against a scratch MariaDB 11.8 schema: both runs succeed, leaving one consent type, one property row and the three columns.
- `check-bdd-test-naming.sh`, `check-security-exception-message-convention.sh`, `check-encoder-null-safety.sh`: PASS.

## Summary by Sourcery

Gate outbound patient SMS on dedicated recorded consent while preserving auditable decisions and safe queue recovery.

New Features:
- Gate patient SMS delivery on a configured, dedicated SMS consent type, allowing sends only for current opt-ins and failing closed for missing, invalid, or opted-out consent.
- Record the consent state and source record used for each SMS transaction as an audit snapshot.

Bug Fixes:
- Release queued SMS claims back to QUEUED without consuming attempts when dispatch-time consent checks fail.
- Handle duplicate live consent records safely by blocking on any opt-out and selecting the latest opt-in otherwise.

Enhancements:
- Replace the deferred SMS consent implementation with database-backed consent resolution through existing consent tables.
- Add explicit consent decision statuses and reason codes for configuration, unknown, opt-out, opt-in, and system-test outcomes.

Build:
- Add an idempotent Flyway migration that seeds the SMS consent type and configuration property and extends SMS transaction auditing fields.

Documentation:
- Document SMS consent decision rules, audit snapshot behavior, worker failure handling, and remaining consent limitations.

Tests:
- Add unit and integration coverage for SMS consent resolution, duplicate records, consent snapshots, and worker recovery behavior.


Ben-Heerema:
@coderabbitai review

